PR #7, opened 2023-04-20
Update standard tree nodes model for BT v4

eholum:
MoveIt Studio v2.0 bumps to BehaviorTree.CPP v4. There are several port names in the standard nodes model that have been updated.

### DIFF
--- a/src/picknik_ur_base_config/objectives/standard_tree_nodes_model.xml
+++ b/src/picknik_ur_base_config/objectives/standard_tree_nodes_model.xml
@@ -35,7 +35,7 @@
                 <p>The ParallelNode executes all its children concurrently, but not in separate threads!</p>
                 <p>Even if this may look similar to ReactiveSequence, this Control Node is the only one that can have multiple children RUNNING at the same time.</p>
                 <p>The Node is completed either when the THRESHOLD_SUCCESS or THRESHOLD_FAILURE number is reached (both configured using ports).</p>
-                <p> If any of the threaholds are reached, and other children are still running, they will be halted.</p>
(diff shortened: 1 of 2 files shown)


src/picknik_ur_base_config/objectives/standard_tree_nodes_model.xml
```diff
@@ -37,8 +37,8 @@
                 <p>The Node is completed either when the THRESHOLD_SUCCESS or THRESHOLD_FAILURE number is reached (both configured using ports).</p>
                 <p> If any of the threaholds are reached, and other children are still running, they will be halted.</p>
```
sea-bass, line 38: not your doing, but noticed this out the corner of my eye

```suggestion
                <p> If any of the threaholds are reached, and other children are still running, they will be halted.</p>
```

sea-bass (dismissed):
Seems to be the case per [the source code](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/5722dd1c46bb23873e5553e5cf7a15c818419dc5/include/behaviortree_cpp/controls/parallel_node.h)!

This needs to be updated in the core Studio repo as well, right?

sea-bass (approved):

Abishalini (dismissed):

Abishalini:
We might want to consider adding `Script` and `Precondition` nodes to the `standard_behavior_tree` XML.
The following ports can also be added to all the nodes - https://www.behaviortree.dev/docs/guides/pre_post_conditions#post-conditions

Abishalini (approved):

PR #11, opened 2023-05-01
Servo settings to avoid collision for mock sim

Abishalini:
Currently, the arm can servo into the table and the only way to come out of collision is to restart Studio.

Reducing the servo speed and increasing the collision check rate helps with displaying a warning and stops collision


EzraBrooks (changes requested):
Shouldn't this be in the mock hardware site config instead of the base config, so as to not regress hardware performance?

sea-bass:
> Shouldn't this be in the mock hardware site config instead of the base config, so as to not regress hardware performance?

The mock hardware isn't a separate config; so the servo settings are shared with hardware the way it's implemented now.

EzraBrooks (approved):
approving per discussion on slack.

PR #12, opened 2023-05-02
Fix name of base config package in xacro file

schornakj:
Fix package name to correctly refer to the base config package in this repo.


sea-bass (approved):
Thanks!

EzraBrooks (approved):
oy, why did we name these differently

MarqRazz (approved):

PR #16, opened 2023-05-11
Add Example Agent call_do_objective.py Scripts

chancecardona:
This just adds a barebones package containing the `call_do_objective.py` scripts for interacting with the MoveIt Studio Agent. This is part of the "Create python tutorial for externally triggering and communicating with Objectives" PRD so that we don't need to reference the file inside of the docker container. 

I'm not sure if I could have done the CMakeLists.txt or the package.xml different. Included are a fairly minimal set of dependencies needed for things to run smoothly and that users could add to without it immediately breaking, but it could possibly be more minimal.

### DIFF
--- a/src/moveit_studio_agent_utils/CMakeLists.txt
+++ b/src/moveit_studio_agent_utils/CMakeLists.txt
@@ -1,0 +1,36 @@
+cmake_minimum_required(VERSION 3.16.3)
+project(moveit_studio_agent_utils)
+
+find_package(ament_cmake REQUIRED)
+find_package(moveit_studio_agent_msgs REQUIRED)
+find_package(moveit_studio_behavior_interface REQUIRED)
+find_package(moveit_studio_behavior_msgs REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(std_msgs REQUIRED)
+
+set(THIS_PACKAGE_INCLUDE_DEPENDS
+    moveit_studio_agent_msgs
+    moveit_studio_behavior_interface
+    moveit_studio_behavior_msgs
+    std_msgs
+)
+
+#############
+## Install ##
+#############
+
+# Install scripts directory
+install(PROGRAMS 
+    scripts/call_do_objective.py 
+    scripts/call_do_objective_waypoint.py
+    DESTINATION lib/${PROJECT_NAME}
+)
+
+
+if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+endif()
+
+ament_export_dependencies(${THIS_PACKAGE_INCLUDE_DEPENDS})
+ament_package()
--- a/src/moveit_studio_agent_utils/package.xml
+++ b/src/moveit_studio_agent_utils/package.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>moveit_studio_agent_utils</name>
+  <version>2.1.0</version>
+  <description>Package containing scripts for interacting with MoveIt Studio Agent</description>
+
+  <maintainer email="chance.cardona@picknik.ai">Chance Cardona</maintainer>
+  <author email="chance.cardona@picknik.ai">Chance Cardona</author>
+
+  <license>BSD-3</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <build_depend>moveit_common</build_depend>
+
+  <depend>moveit_studio_agent_msgs</depend>
+  <depend>moveit_studio_behavior_interface</depend>
+  <depend>moveit_studio_behavior_msgs</depend>
+  <depend>std_msgs</depend>
+
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>ament_clang_format</test_depend>
+  <test_depend>ament_clang_tidy</test_depend>
+  <test_depend>ament_cmake_copyright</test_depend>
+  <test_depend>ament_cmake_lint_cmake</test_depend>
+  <test_depend>picknik_ament_copyright</test_depend>
+
+  <export>
+    <build_type>ament_cmake</build_type>
+  </export>
+</package>
--- a/src/moveit_studio_agent_utils/scripts/call_do_objective.py
+++ b/src/moveit_studio_agent_utils/scripts/call_do_objective.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+
+# Copyright 2022 PickNik Inc.
+# All rights reserved.
+#
+# Unauthorized copying of this code base via any medium is strictly prohibited.
+# Proprietary and confidential.
+
+import rclpy
+from rclpy.action import ActionClient
+from rclpy.node import Node
+import sys
+
+from moveit_studio_agent_msgs.action import DoObjectiveSequence
+
+
+class DoObjectiveSequenceClient(Node):
+    def __init__(self):
+        super().__init__("DoObjectiveSequence")
+        self._action_client = ActionClient(self, DoObjectiveSequence, "do_objective")
+
+    def send_goal(self, objective_name):
+        goal_msg = DoObjectiveSequence.Goal()
+        goal_msg.objective_name = objective_name
+        self._action_client.wait_for_server()
+        result = self._action_client.send_goal_async(goal_msg)
+        return result
+
+
+def main(args=None):
+    if len(sys.argv) < 2:
+        print(
+            "usage: ros2 run moveit_studio_behavior call_do_objective.py 'Objective Name'"
+        )
+    else:
+        rclpy.init(args=args)
+
+        client = DoObjectiveSequenceClient()
+
+        future = client.send_goal(sys.argv[1])
+
+        rclpy.spin_until_future_complete(client, future)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/moveit_studio_agent_utils/scripts/call_do_objective_waypoint.py
+++ b/src/moveit_studio_agent_utils/scripts/call_do_objective_waypoint.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+
+# Copyright 2022 PickNik Inc.
+# All rights reserved.
+#
+# Unauthorized copying of this code base via any medium is strictly prohibited.
+# Proprietary and confidential.
+
+import rclpy
+from rclpy.action import ActionClient
+from rclpy.node import Node
+import sys
+
+from moveit_studio_agent_msgs.action import DoObjectiveSequence
+from moveit_studio_behavior_msgs.msg import BehaviorParameter, BehaviorParameterDescription
+
+class DoObjectiveSequenceClient(Node):
+    def __init__(self):
+        super().__init__("DoObjectiveSequence")
+        self._action_client = ActionClient(self, DoObjectiveSequence, "do_objective")
+
+    def send_goal(self, objective_name, waypoint_name="Behind"):
+        goal_msg = DoObjectiveSequence.Goal()
+        goal_msg.objective_name = objective_name
+
+        behavior_parameter = BehaviorParameter()
+        behavior_parameter.behavior_namespaces.append("move_to_joint_state")
+        behavior_parameter.description.name = "waypoint_name"
+        behavior_parameter.description.type = BehaviorParameterDescription.TYPE_STRING
+        behavior_parameter.string_value = waypoint_name
+        goal_msg.parameter_overrides = [behavior_parameter]
+
+        self._action_client.wait_for_server()
+        result = self._action_client.send_goal_async(goal_msg)
+        return result
+
+
+def main(args=None):
+    if len(sys.argv) < 2:
+        print(
+            "usage: ros2 run moveit_studio_behavior call_do_objective.py 'Objective Name'"
+        )
+    else:
+        rclpy.init(args=args)
+
+        client = DoObjectiveSequenceClient()
+
+        future = client.send_goal(sys.argv[1], sys.argv[2])
+
+        rclpy.spin_until_future_complete(client, future)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/moveit_studio_agent_utils/scripts/call_do_objective_waypoint.py
+++ b/src/moveit_studio_agent_utils/scripts/call_do_objective_waypoint.py
@@ -12,16 +12,19 @@ from rclpy.node import Node
 import sys
 
 from moveit_studio_agent_msgs.action import DoObjectiveSequence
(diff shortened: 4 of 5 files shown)


src/moveit_studio_agent_utils/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+find_package(ament_cmake REQUIRED)
+find_package(moveit_studio_agent_msgs REQUIRED)
+find_package(moveit_studio_behavior_interface REQUIRED)
+find_package(moveit_studio_behavior_msgs REQUIRED)
+find_package(pluginlib REQUIRED)
+find_package(std_msgs REQUIRED)
+
+set(THIS_PACKAGE_INCLUDE_DEPENDS
+    moveit_studio_agent_msgs
+    moveit_studio_behavior_interface
```
eholum, line 13: Are all the deps here needed or just the agent msgs?

src/moveit_studio_agent_utils/scripts/call_do_objective_waypoint.py
```diff
@@ -0,0 +1,80 @@
...
+from rclpy.node import Node
+import sys
+
+from moveit_studio_agent_msgs.action import DoObjectiveSequence
+from moveit_studio_behavior_msgs.msg import (
+    BehaviorParameter, 
+    BehaviorParameterDescription,
+)
+
+class DoObjectiveSequenceClient(Node):
```
eholum, line 43: Add class and method comments here and to the `send_goal` function below.

src/moveit_studio_agent_utils/scripts/call_do_objective.py
```diff
@@ -0,0 +1,69 @@
...
+
+import rclpy
+from rclpy.action import ActionClient
+from rclpy.node import Node
+import sys
+
+from moveit_studio_agent_msgs.action import DoObjectiveSequence
+
+
+class DoObjectiveSequenceClient(Node):
```
eholum, line 40: Add class and method comments here and to the `send_goal` function below.

src/moveit_studio_agent_utils/scripts/call_do_objective.py
```diff
@@ -0,0 +1,69 @@
...
+            "usage: ros2 run moveit_studio_behavior call_do_objective.py 'Objective Name'"
+        )
+    else:
+        rclpy.init(args=args)
+
+        client = DoObjectiveSequenceClient()
+
+        future = client.send_goal(sys.argv[1])
+
+        rclpy.spin_until_future_complete(client, future)
```
eholum, line 65: Maybe print something on success or failure? Could check for a non-empty error message and string and log it if the objective failed.

eholum (changes requested):
Thanks for putting this together! We should hold this until the 2.1 release goes live for the docs link, right?

src/moveit_studio_agent_utils/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 3.16.3)
+project(moveit_studio_agent_utils)
```
sea-bass, line 2: I would call this package `moveit_studio_agent_examples`, perhaps?

src/moveit_studio_agent_utils/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+find_package(ament_cmake REQUIRED)
+find_package(moveit_studio_agent_msgs REQUIRED)
+find_package(moveit_studio_behavior_msgs REQUIRED)
+
+set(THIS_PACKAGE_INCLUDE_DEPENDS
```
sea-bass, line 8: For completeness, let's put a comment here saying these are packages needed to link against the necessary MoveIt Studio ROS interface definitions. Interface being the official ROS 2 term for "messages + services + actions"

src/moveit_studio_agent_utils/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+#############
+
+# Install scripts directory
+install(PROGRAMS 
+    scripts/call_do_objective.py 
+    scripts/call_do_objective_waypoint.py
+    DESTINATION lib/${PROJECT_NAME}
+)
+
+
```
sea-bass, line 24: Remove extra line

src/moveit_studio_agent_utils/package.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>moveit_studio_agent_utils</name>
+  <version>2.1.0</version>
+  <description>Package containing scripts for interacting with MoveIt Studio Agent</description>
+
+  <maintainer email="chance.cardona@picknik.ai">Chance Cardona</maintainer>
+  <author email="chance.cardona@picknik.ai">Chance Cardona</author>
+
+  <license>BSD-3</license>
```
sea-bass, line 10: for consistency

```suggestion
  <license>BSD-3-Clause</license>
```

src/moveit_studio_agent_utils/package.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>moveit_studio_agent_utils</name>
+  <version>2.1.0</version>
```
sea-bass, line 4: This is going in for 2.2.0, right? So make it 2.2.0

src/moveit_studio_agent_utils/package.xml
```diff
@@ -0,0 +1,30 @@
...
+  <description>Package containing scripts for interacting with MoveIt Studio Agent</description>
+
+  <maintainer email="chance.cardona@picknik.ai">Chance Cardona</maintainer>
+  <author email="chance.cardona@picknik.ai">Chance Cardona</author>
+
+  <license>BSD-3</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <build_depend>moveit_common</build_depend>
```
sea-bass, line 14: Is this necessary?

src/moveit_studio_agent_utils/scripts/call_do_objective.py
```diff
@@ -0,0 +1,82 @@
...
+    if len(sys.argv) < 2:
+        print(
+            "usage: ros2 run moveit_studio_agent_utils call_do_objective.py 'Objective Name'"
+        )
+    else:
+        rclpy.init(args=args)
+
+        client = DoObjectiveSequenceClient()
+
+        future = client.send_goal(sys.argv[1])
```
sea-bass, line 76: For readability, I'd do something like:

```suggestion
        objective_name = sys.argv[1]
        future = client.send_goal(objective_name)
```

and same for the waypoint name arg in the other script

sea-bass (commented):
Thanks for adding pre-commit to this repo!

Things look good -- I didn't have a whole lot more to add outside @eholum's comments, which were great.

Definitely looking forward to something around cancellation as well.

src/moveit_studio_agent_utils/CMakeLists.txt
chancecardona: They definitely were not all needed. Before I had run into some errrors but this time it all just worked, so now it's the reduced set.

src/moveit_studio_agent_utils/package.xml
chancecardona: It was not, good catch! It can be hard finding all of what's still leftover sometimes with these files I swear.

src/moveit_studio_agent_utils/CMakeLists.txt
chancecardona: Updated, thanks for the suggestion.

src/moveit_studio_agent_utils/scripts/call_do_objective.py
chancecardona: It required 2 more callbacks, but I do think it makes the script behave a good deal nicer.

chancecardona:
> Definitely looking forward to something around cancellation as well.
@sea-bass I have added a script to cancel the objective on a timer after 2 seconds. 

I also have a script that could do this standalone, but it needs to cancel a goal through the goal's `ClientGoalHandle`, of which I'm unsure how to either get from the Action Server / propagate between nodes. Is there some easy way to do this that I'm missing (like a studio topic or method)? Maybe we could set a parameter on the DoObjectiveSequence node and get it from there? Either way, I think the current example should be good / may be the simplest.

src/moveit_studio_agent_examples/scripts/call_do_objective.py
```diff
@@ -0,0 +1,111 @@
...
+
+        self._goal_handle = goal_handle
+        self.get_logger().info("Goal accepted...")
+
+        get_result_future = goal_handle.get_result_async()
+        get_result_future.add_done_callback(self.get_result_callback)
+
+    def get_result_callback(self, future):
+        result = future.result().result
+        if result.error_code.val == 1:
```
sea-bass, line 80: Instead of using the literal `1` can you use the actual enum from https://github.com/ros-planning/moveit_msgs/blob/ros2/msg/MoveItErrorCodes.msg?

It should be something like

```python
from moveit_msgs.msg import MoveItErrorCodes
...
if result.error_code.val == MoveItErrorCodes.SUCCESS
```

Same with the other scripts.

src/moveit_studio_agent_examples/scripts/call_do_objective.py
```diff
@@ -0,0 +1,111 @@
...
+from rclpy.node import Node
+import sys
+
+from moveit_studio_agent_msgs.action import DoObjectiveSequence
+
+
+class DoObjectiveSequenceClient(Node):
+    """
+    ROS 2 node that acts as an Action Client for the MoveIt Studio Agent's Objective Server
+    """
```
sea-bass, line 43: I think for each of these separate files, the comment should be more specific to that example, instead of all being the same.

For example, this one can stay as is but the other scripts should say what additional you do -- like, "adds a parameter override for waypoint names" or whatever.

sea-bass (commented):
Looking great!

I think for your question, you're right -- I didn't think about the fact that you'd need a goal handle to cancel it.

I know this is kind of a 180 from the original comments, but is it worth possibly adding a boolean flag to the original script to enable canceling, instead of having one that largely copy-pastes the original one?

I think it's OK if such a parameter is hardcoded, but you can make it a command-line arg as well if you find it beneficial. Either with `sys.argv` or `argparse` if you wanna get fancy.

src/moveit_studio_agent_examples/scripts/call_do_objective.py
chancecardona: Ok, updated to have an additional line. Can always add more, not sure if it'd be helpful for any of the callbacks or anything too.

chancecardona:
> I think for your question, you're right -- I didn't think about the fact that you'd need a goal handle to cancel it.

Yeah it is kind of weird about the goal handle stuff. If this [feature request ](https://github.com/PickNikRobotics/moveit_studio_ws/pull/16) would be useful I can always attempt to implement it.

> I know this is kind of a 180 from the original comments, but is it worth possibly adding a boolean flag to the original script to enable canceling, instead of having one that largely copy-pastes the original one?
> 
> I think it's OK if such a parameter is hardcoded, but you can make it a command-line arg as well if you find it beneficial. Either with `sys.argv` or `argparse` if you wanna get fancy.

Ok, I have added argparse and adapted the examples so all have a `--cancel` optional toggle to cancel the objective after 2 seconds.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,73 @@
+# To use:
```
eholum, line 1: Yes! Thank your for adding this!

src/moveit_studio_agent_examples/package.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>moveit_studio_agent_examples</name>
+  <version>2.2.0</version>
```
eholum, line 4:
```suggestion
  <version>2.1.0</version>
```

src/moveit_studio_agent_examples/scripts/call_do_objective.py
```diff
@@ -0,0 +1,144 @@
...
+from rclpy.node import Node
+import sys
+
+from moveit_msgs.msg import MoveItErrorCodes
+from moveit_studio_agent_msgs.action import DoObjectiveSequence
+
+
+class DoObjectiveSequenceClient(Node):
+    """
+    ROS 2 node that acts as an Action Client for the MoveIt Studio Agent's Objective Server
```
eholum, line 43:
```suggestion
    ROS 2 node that acts as an Action Client for MoveIt Studio's Objective Server.
```

src/moveit_studio_agent_examples/scripts/call_do_objective_waypoint.py
```diff
@@ -0,0 +1,157 @@
...
+from moveit_studio_agent_msgs.action import DoObjectiveSequence
+from moveit_studio_behavior_msgs.msg import (
+    BehaviorParameter,
+    BehaviorParameterDescription,
+)
+
+
+class DoObjectiveSequenceClient(Node):
+    """
+    ROS 2 node that acts as an Action Client for the MoveIt Studio Agent's Objective Server.
```
eholum, line 47:
```suggestion
    ROS 2 node that acts as an Action Client for MoveIt Studio's Objective Server.
```

src/moveit_studio_agent_examples/scripts/call_do_objective_waypoint.py
```diff
@@ -0,0 +1,157 @@
...
+        behavior_parameter.description.type = BehaviorParameterDescription.TYPE_STRING
+        behavior_parameter.string_value = waypoint_name
+        goal_msg.parameter_overrides = [behavior_parameter]
+
+        self._action_client.wait_for_server()
+        self._send_goal_future = self._action_client.send_goal_async(goal_msg)
+        self._send_goal_future.add_done_callback(self.goal_response_callback)
+        return self._send_goal_future
+
+    def goal_response_callback(self, future):
```
eholum, line 81: Oof. Yes action servers have a lot of overhead but I think this is a great example of how to use their callbacks from a script.

eholum (commented):
Looks really good! Some minor nit-picky things but I think this is a great example to have. Thanks for putting it together!

src/moveit_studio_agent_examples/scripts/call_do_objective_waypoint.py
chancecardona: I agree, I find the callback part added is the part I (as a user) would actually want to see an example of how to do on the whole `Action Client` side of things. I debated passing the `cancel` argument to a callback function by passing a lambda where that's set (which appears to be the proper ros2 way of doing it) so they could see that, but figured for a quick example it may be simpler to just store it in the class and use `self.cancel` instead.

sea-bass (approved):

PR #19, opened 2023-05-26
Fix Copyright Statements for Ament Test

eholum:
So that ament tooling can verify our statements and license type.

Verify that running the test builder passes with these changes.


sea-bass (approved):
LGTM!

image

image

PR #21, opened 2023-05-29
Add Waypoint Cycling Objective

bgill92:
# Description
Resolves PickNikRobotics/moveit_studio#3161
This PR adds an objective that makes the arm move between two waypoints until failure.

# Testing
This objective was tested with both the mock hardware and gazebo sim using Studio version 2.1.0.


sea-bass (approved):
LGTM!

EzraBrooks (approved):
I trust the sea bass and this is getting released

PR #22, opened 2023-05-30
Add Admittance Control to Open Lever Handle Door

chancecardona:
Studio side PR for this change here: https://github.com/PickNikRobotics/moveit_studio/pull/3198. This just handles changing the files here that need to be changed. Please see other PR for testing/context.

~~Draft until https://github.com/PickNikRobotics/moveit_studio/pull/3157 is merged.~~


sea-bass (dismissed):

src/picknik_ur_site_config/objectives/open_lever_handle_door.xml
```diff
@@ -5,6 +5,10 @@
         <Control ID="Sequence" name="Setup">
             <Action ID="LoadObjectiveParameters"
                  config_file_name="open_lever_handle_door_config.yaml" parameters="{parameters}" />
+            <Decorator ID="ForceSuccess">
```
sea-bass, line 8: Is this so the Objective still runs in Gazebo where there is no admittance controller? If so, might be worth dropping a comment.

chancecardona: Yeah this is why. It also has some of the odd behavior (same as the Tare FTS service) where if planning fails then it also complains it can't find the correct controller. In the future we should try to fix that for both of these. In the meantime I'll just update the comments.

chancecardona: Updated comments. I think it should be included so it works on sim (even if the behavior otherwise fails for different reasons such as our scene not having what it needs). Issue so we can fix this properly.

src/picknik_ur_site_config/objectives/open_cabinet_door_affordance.xml
```diff
@@ -3,13 +3,17 @@
...
         <Control ID="Sequence" name="Setup">
             <!-- Tare the force torque sensor -->
+            <!-- Wrap in a ForceSuccess as Sim doesn't have the zero_ftsensor service -->
+            <!-- See https://github.com/PickNikRobotics/moveit_studio/issues/3406 -->
             <Decorator ID="ForceSuccess">
                 <Action ID="CallTriggerService" service_name="/io_and_status_controller/zero_ftsensor" />
             </Decorator>
             <!-- Set admittance control parameters at the start of the Objective -->
-            <Action ID="UpdateAdmittanceController" config_file_name="open_door_admittance_controller_config.yaml"/>
+            <Decorator ID="ForceSuccess">
```
sea-bass, line 12: Maybe I pointed at the wrong decorator, but I meant to add a comment to this one as well. This one needs to be force success'd because sim doesn't have admittance controller loaded.

src/picknik_ur_site_config/objectives/open_cabinet_door_ml.xml
```diff
@@ -4,12 +4,17 @@
...
+            <!-- Wrap in a ForceSuccess as Sim doesn't have the zero_ftsensor service -->
+            <!-- This should eventually get overridden for non sim site configs to not ForceSuccess -->
+            <!-- See https://github.com/PickNikRobotics/moveit_studio/issues/3406 -->
             <Decorator ID="ForceSuccess">
                 <Action ID="CallTriggerService" 
                     service_name="/io_and_status_controller/zero_ftsensor" />
             </Decorator>
-            <Action ID="UpdateAdmittanceController"
-                config_file_name="open_door_admittance_controller_config.yaml" />
+            <Decorator ID="ForceSuccess">
```
sea-bass, line 14: and same here

src/picknik_ur_site_config/objectives/open_cabinet_door_affordance.xml
chancecardona: Gotcha, updated others to follow suit.

sea-bass (approved):

PR #25, opened 2023-06-01
Modify objectives to use Get String/Pose/JointState From User behavior

Abishalini:
Corresponding PR - https://github.com/PickNikRobotics/moveit_studio/pull/3118


src/picknik_ur_base_config/objectives/interpolate_to_joint_state.xml
```diff
@@ -2,7 +2,11 @@
 <root BTCPP_format="4" main_tree_to_execute="Interpolate to Joint State">
     <BehaviorTree ID="Interpolate to Joint State" _description="Move to a specified joint state">
         <Control ID="Sequence" name="root">
-            <SubTree ID="InterpolateToJointStateWaypoint" target_joint_state="{interpolate_to_joint_state.target_joint_state}"/>
+            <Fallback name="wait_for_approval_if_user_available">
+            <Action ID="GetJointStateFromUser" parameter_name="interpolate_to_joint_state.target_joint_state" parameter_value="{target_joint_state}" />
+                <Action ID="LogMessage" message="This objective is used by joint sliders in the Joints tab." log_level="error" />
+            </Fallback>
```
schornakj, line 8: A few things to mention here:

- Fix the indentation for the XML.
- Write the `Fallback` node in the same way  as the objective editor in the UI serializes it, which is `Control ID="Fallback" name =...`.
- The fallback condition needs to be a `Sequence` where the first child is `LogMessage` and the second child is `AlwaysFailure` (or, put `LogMessage` under a `ForceFailure` decorator, but I find that using `AlwaysFailure` is more readable because it shows the steps clearly). What happens right now when we run this objective as currently written is that `GetJointStateFromUser` runs and fails because no user input was provided, which triggers the fallback branch. Since `LogMessage` succeeds when it runs, it makes the fallback control succeed, so the objective keeps running and the subtree that does the motion planning and execution ends up being run even though it should not.

src/picknik_ur_base_config/objectives/move_to_known_pose.xml
```diff
@@ -2,7 +2,13 @@
 <root BTCPP_format="4" main_tree_to_execute="Move to Known Pose">
     <BehaviorTree ID="Move to Known Pose" _description="Move to a known pose">
         <Control ID="Sequence" name="root">
-            <SubTree ID="MoveToWaypoint" waypoint_name="{move_to_known_pose.target_state_name}"/>
+            <!-- When the user clicks on different pose buttons on Endpoint tab, the frontend sets move_to_known_pose.target_state_name parameter value-->
+            <Fallback name="log_error_if_parameter_not_set">
+                <Action ID="GetStringFromUser" parameter_name="move_to_known_pose.target_state_name" parameter_value="{target_state_name}"/>
+                <Action ID="LogMessage" message="This objective is used by Move To Pose buttons in the Endpoint tab." log_level="error" />
+            </Fallback>
```
schornakj, line 9: Same issue here as I described in my other comment.

src/picknik_ur_base_config/objectives/move_to_known_pose.xml
```diff
@@ -2,7 +2,13 @@
...
+            <Action ID="GetStringFromUser" parameter_name="move_to_known_pose.target_state_name" parameter_value="{target_state_name}"/>
```
schornakj, line 10: This line seems duplicated and should be deleted.

schornakj (changes requested):

schornakj (approved):

src/picknik_ur_base_config/objectives/interpolate_to_joint_state.xml
```diff
@@ -2,7 +2,14 @@
 <root BTCPP_format="4" main_tree_to_execute="Interpolate to Joint State">
     <BehaviorTree ID="Interpolate to Joint State" _description="Move to a specified joint state">
         <Control ID="Sequence" name="root">
-            <SubTree ID="InterpolateToJointStateWaypoint" target_joint_state="{interpolate_to_joint_state.target_joint_state}"/>
+            <Control ID="Fallback" name="log_error_if_parameter_not_set">
+                <Action ID="GetJointStateFromUser" parameter_name="interpolate_to_joint_state.target_joint_state" parameter_value="{target_joint_state}" />
+                <Control ID="Sequence">
+                    <Action ID="LogMessage" message="This objective is meant to be used by joint sliders in the Joints tab. Try it out there!" log_level="error" />
```
sea-bass, line 8:
```suggestion
                    <Action ID="LogMessage" message="This Objective is meant to be used by joint sliders in the Joints tab. Try it out there!" log_level="error" />
```

src/picknik_ur_base_config/objectives/move_to_known_pose.xml
```diff
@@ -2,7 +2,15 @@
 <root BTCPP_format="4" main_tree_to_execute="Move to Known Pose">
     <BehaviorTree ID="Move to Known Pose" _description="Move to a known pose">
         <Control ID="Sequence" name="root">
-            <SubTree ID="MoveToWaypoint" waypoint_name="{move_to_known_pose.target_state_name}"/>
+            <!-- When the user clicks on different pose buttons on Endpoint tab, the frontend sets move_to_known_pose.target_state_name parameter value-->
+            <Control ID="Fallback" name="log_error_if_parameter_not_set">
+                <Action ID="GetStringFromUser" parameter_name="move_to_known_pose.target_state_name" parameter_value="{target_state_name}"/>
+                <Control ID="Sequence">
+                    <Action ID="LogMessage" message="This objective is meant to be used by Move To Pose buttons in the Endpoint tab. Try it out there!" log_level="error" />
```
sea-bass, line 9:
```suggestion
                    <Action ID="LogMessage" message="This Objective is meant to be used by Move To Pose buttons in the Endpoint tab. Try it out there!" log_level="error" />
```

sea-bass (commented):
Objective needs to be capitalized in those log messages, but otherwise LGTM!

sea-bass (approved):

schornakj (approved):

PR #32, opened 2023-06-08
Clean up config and launch files to reduce duplication

sea-bass:
WIP: Associated with a cleanup PR to MoveIt Studio


src/picknik_ur_base_config/config/base_config.yaml
```diff
@@ -38,29 +38,23 @@ hardware:
...
     package: "picknik_ur_base_config"
     path: "launch/robot_drivers_to_persist.launch.py"
-  robot_driver_restart_launch_file:
-    package: "picknik_ur_base_config"
-    path: "launch/robot_drivers_to_restart.launch.py"
   hardware_launch_file:
-    package: "picknik_ur_base_config"
-    path: "launch/hardware.launch.py"
+    package: "moveit_studio_agent"
+    path: "launch/blank.launch.py"
```
sea-bass, line 47: note: while this is not necessary (it's what we will fall back on if you don't specify this entry), I wanted to keep it in for completeness

src/picknik_ur_base_config/config/base_config.yaml
```diff
@@ -38,29 +38,23 @@ hardware:
   ip: "!!! Please set the IP address of the robot"
 
   # Specify additional launch files for running the robot with real hardware.
-  # [Required]
+  # [Optional, defaults to a blank launch file if not specified]
```
chancecardona, line 41: Are we planning on using these tags for anything automated? I only ask because if so the second part would be better served as a comment. If these are just comment tags anyways I think this is fine.

sea-bass: I don't think they're used for anything automated, they were there just to pop out for users reading the config files... but I see your point for consistency.

src/picknik_ur_base_config/config/base_config.yaml
Abishalini: Should we add a blank file? Wouldn't Studio crash when it tries to find the file that doesn't exist?

Abishalini (approved):
Did not test but the changes look good to me

src/picknik_ur_base_config/config/base_config.yaml
sea-bass: No, actually! The associated PR to MoveIt Studio will automatically replace any missing launch file entries with the above blank launch file.

chancecardona (approved):
Tested on Gazebo and site_config. Works for me!

PR #33, opened 2023-06-13
Increase UR5e wrist3 joint position limits to use full mechanical range

schornakj:


sea-bass (approved):
LGTM, but is there some motivation for expanding this specific limit?

schornakj:
Increasing this limit helps the robot plan shorter paths to some poses during objectives like Pick and Place Object, since now it can just rotate the wrist a little bit further rather than needing to perform a big joint flip motion.

eholum (approved):

PR #45, opened 2023-06-22
Update Pinch Link

chancecardona:
This PR should resolve https://github.com/PickNikRobotics/moveit_studio/issues/2990. This lengthens the pinch link to avoid the p-stop issues, and adds the tool_changer_link to the excluded collisions as it is not part of UR's driver that causes this p-stop, and not including it causes planning weirdness sometimes.


sea-bass (dismissed):
Was tested on hardware, right?

chancecardona:
Tested this on hardware, but it didn't seem to resolve the issue. Revising and retesting, will update when I have a good candidate.

cmelliott2 (approved):

PR #48, opened 2023-06-22
Add "Joint Diagnostic" Objective

bgill92:
Resolves https://github.com/PickNikRobotics/moveit_studio/issues/3694
Resolves https://github.com/PickNikRobotics/moveit_studio/issues/3685

This PR adds an objective which first moves the arm to the home pose, and then cycles between the min and max limits of the Wrist 2 joint. 

**Updated**

To test, launch studio, build the workspace, and test the "Joint Diagnostic" Objective with both the mock hardware, Gazebo simulation.

Additionally, make sure that the "Joint Diagnostic" Objective does not show up when using hardware.


sea-bass (commented):
Can you resolve the formatting comments?

Also, should we remove the Objective introduced in https://github.com/PickNikRobotics/moveit_studio_ws/pull/21?

Lastly, are there are doc changes required in the main repo, or have we not yet documented the joint cycling stuff?

nbbrooks:
Would it be possible to change the joint state definitions on the hardware platforms we deploy to to be a smaller range (to avoid damaging the camera, etc)? But mock_sim/Gz would still do the full range?

sea-bass:
Re: @nbbrooks ' request to have different software vs. hardware limits.

You'll need the Objective XML to be in the `site` and `gazebo` configs in this package, but NOT IN the `base`... and then all the other hardware-specific configs can similarly have their own XMLs.

I'm working on making overrides nicer, but it's not ready at the moment. Sorry.

src/picknik_ur_base_config/waypoints/ur5e_waypoints.yaml
```diff
@@ -181,3 +181,25 @@
...
+    effort: []
+- name: Wrist 2 Max
+  joint_state:
+    header:
+      frame_id: world
+      stamp:
+        sec: 0
+        nanosec: 0
+    name: [shoulder_pan_joint, shoulder_lift_joint, elbow_joint, wrist_1_joint, wrist_2_joint, wrist_3_joint]
+    position: [-2.2335834056138996e-05, -0.7800858658974525, -0.5100288585397881, -2.399985082258936, 3.141592653589793, 2.282092208042741e-05]
```
chancecardona, line 203: nit: This and the gazebo waypoints should have consistent formatting

sea-bass (approved):
This looks good!

chancecardona (approved):
FINALLY was able to test this, lol. This works for me! Tested on 002 (nothing showed up as intended) and on mock (picknik_ur_site_config), which worked.

sushanPicknik:
I pulled main but i am not seeing this objective. I might have missed a step, was just trying to get a jump on testing 2.4
image

sushanPicknik:
I was prompted that my file with waypoints would be deleted and it did pull 2gb so was assuming this should be the right objectives list

nbbrooks:
Replying offline so someone else may have answered already, but if you are
looking for this on HW it should not be there. There wasn't a great example
range for HW I felt comfortable running on ours/others with regard to large
motions and cable wrap. This objective is meant to add more functionality
to HW and be easily adaptable for specific customers.

PR #53, opened 2023-06-27
Make UR base config work with any UR arm

Abishalini:
An attempt to resolve - https://github.com/PickNikRobotics/moveit_studio/issues/3275


src/picknik_ur_base_config/config/base_config.yaml
```diff
@@ -90,17 +90,17 @@ hardware:
       # To use parameters from a different package, place them in a config/ROBOT_NAME/ directory,
       # replace ROBOT_NAME with the value used for hardware.type in this file.
       - kinematics_parameters_file:
-      # Load default_kinematics.yaml from ur_description/config/ur5e
+      # Load default_kinematics.yaml from ur_description/config/<ur_type>
           package: "ur_description"
-          path: "config/ur5e/default_kinematics.yaml"
+          path: "config/%>> hardware.type <<%/default_kinematics.yaml"
```
sea-bass, line 95: Does this `%>> blah <<%` syntax actually work? I wasn't aware of this, nor do I see it in the source?

src/picknik_ur_base_config/config/base_config.yaml
```diff
@@ -223,4 +223,4 @@ objectives:
   # [Required]
   waypoints_file:
     package_name: "picknik_ur_base_config"
-    relative_path: "waypoints/ur5e_waypoints.yaml"
+    relative_path: "waypoints/ur_waypoints.yaml"
```
MarqRazz, line 226: I worry about letting someone choose the UR model and giving them access to waypoints that were specific to a UR5e. What do you think about stripping this file back to a single pose that works on all models (like a zero pose)?

schornakj (approved):
I tried this out on the UR5 and it loads the correct robot config data now.

sea-bass (approved):

PR #56, opened 2023-07-06
Update favorite objectives for the GZ simulation

eholum:


EzraBrooks (approved):
I guess it's fine 🤔

sea-bass (approved):
I'll be adding a way to override such objectives in some upcoming work... however, it will involve copy-pasting the XMLs to the Gazebo config because they need the "favorite" to be True :/

eholum:
FWIW all of the remaining objectives will fail in Mock Simulation because of a missing camera - excepting for looping pick and place and snapshot, which fail because they can't find specified images and point clouds on the specified topics, respectively.

PR #58, opened 2023-07-06
Reorganize for generic config package inheritance

sea-bass:
This PR reorganizes the config packages since we're adding generic inheritance.

Now the inheritance chain is: `picknik_ur_base_config` -> `picknik_ur_site_config` -> `picknik_ur_gazebo_config`


src/picknik_ur_site_config/objectives/open_cabinet_door_ml.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0"?>
+<root BTCPP_format="4" main_tree_to_execute="Open Cabinet Door ML">
```
sea-bass, line 2: NOTE: These ML Objectives won't work, but I also don't want to copy paste them in all the other configurations.

Maybe we can add something that checks whether the ML server is available, and if not we can enter a fallback that prints an informative log message?

eholum: Yep, as noted you get a nice "ML Action server is not available error" if it is not available. I think that's sufficient.

eholum (approved):

PR #60, opened 2023-07-07
Remove old links, point users to public docs by default

eholum:
After the major public documentation reorg many of these links have gone stale. We should default to the getting started guides to reduce future overhead.


cmelliott2 (approved):
LGTM 👍

PR #63, opened 2023-07-11
Add parameters to toggle camera launching on hardware

sea-bass:
This adds the `launch_cameras` parameter to the example config packages in this repo.

### DIFF
--- a/src/picknik_ur_gazebo_config/config/site_config.yaml
+++ b/src/picknik_ur_gazebo_config/config/site_config.yaml
@@ -31,6 +31,9 @@ hardware:
   # This should be false if you are launching the robot state publisher as part of drivers.
   launch_robot_state_publisher: True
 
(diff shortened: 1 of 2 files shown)


src/picknik_ur_gazebo_config/config/site_config.yaml
```diff
@@ -31,6 +31,9 @@ hardware:
   # This should be false if you are launching the robot state publisher as part of drivers.
   launch_robot_state_publisher: True
 
+  # If the MoveIt Studio Agent should launch cameras when running on hardware.
+  launch_cameras: True
+
   # If the MoveIt Studio Agent should launch cameras when simulated.
   launch_cameras_when_simulated: True
```
MarqRazz, line 38: Suggestion: if we are going to add `launch_cameras` as an option should we consider refactoring `launch_cameras_when_simulated`? Technically it's not launching cameras but telling the UI to subscribe to the simulated topics.

PR #66, opened 2023-07-14
Fixes for the Mock Sim and Gazebo Simulated Configurations

sea-bass:
Some changes for v2.4:

* The Gazebo config has lingering objective overrides from the days of 2 layer limitations. Inspect Surface and Take Snapshot XMLs don't need to be there. Inheriting from the site config fixes the issue.
* Making better choices for favorite objectives.
* Updating the start pose for the UR5 to not immediately servo into the wrist-shoulder singularity.


cmelliott2:
Why is only 1 check running in CI for this PR?

sea-bass:
> Why is only 1 check running in CI for this PR?

It's the `moveit_studio_ws` dawg

cmelliott2:
> > Why is only 1 check running in CI for this PR?
> 
> It's the `moveit_studio_ws` dawg

Oops. My bad

eholum:
@cmelliott2 @sea-bass I've cleaned up a few other issues identified as part of v2.4 release testing. Please take a look at let me know, as I am loathe to approve a PR that I have so many changes on.

nbbrooks:
> Updating the start pose for the UR5 to not immediately servo into the wrist-shoulder singularity.

If we are talking about #3945 here, just making sure people are aware we need to move the whole environment forward (or the robot backward) so we can do everything with the wrist "in front of" the base, otherwise we're going to keep running into this sigularity. Just changing starting joint positions won't help. Just want to make sure this is what we mean by "update start pose"

sea-bass:
@nbbrooks could you expand on what situations get you near those singularities? The blocks I am seeing in the Gazebo world are here relative to the robot:

image

sea-bass:
@eholum your changes LGTM -- but I can't approve my own PR

eholum:
> > Updating the start pose for the UR5 to not immediately servo into the wrist-shoulder singularity.
> 
> If we are talking about #3945 here, just making sure people are aware we need to move the whole environment forward (or the robot backward) so we can do everything with the wrist "in front of" the base, otherwise we're going to keep running into this sigularity. Just changing starting joint positions won't help. Just want to make sure this is what we mean by "update start pose"

This only resolves the issue that users see when they start mock sim and immediately drive the end effector into the singularity in front of the robot. Likewise the GZ sim will now start with the wrist camera pointed at the objects on the table.

eholum (approved):

nbbrooks:
I'm talking about elbow singularity it passes through whenever the wrist moves from behind the base of the arm to in front of it. You run into it in gz (and maybe mock sim) if you hit + as it does this. We should try and start the robot with the wrist in front of the base and position things in the environment such that when people teleport they wouldn't have a need to move the wrist behind the base

PR #70, opened 2023-07-20
Add missing dependency on moveit_planners_stomp package

schornakj:
`picknik_ur_base_config` loads the `StompPlanner` plugin as one of the available motion planning pipelines, so it needs to have an `exec_depend` on `moveit_planners_stomp`.

I think this only causes problems for the rare user who is building the entire underlay workspace from source (i.e., me), since the Studio Docker image builds all packages in the MoveIt2 repo every time regardless of which overlay packages depend on them.


sea-bass (approved):
nice catch!

sjahr (approved):

PR #72, opened 2023-07-21
Add Machine Tending Mock Hardware Configuration

kineticsystem:
This PR adds a new config package named `picknik_ur_mock_hw_config` which contains a machine tending scene. This will now be the new default when installing.

https://github.com/PickNikRobotics/moveit_studio_ws/assets/4603398/a4227007-7a9e-482d-bf68-ffbd93294330

Note that `picknik_ur_site_config` will still remain for situations with less compute power, and also for general integration testing on the MoveIt Studio side.

The hierarchy now is 

```mermaid
graph TB
Base[picknik_ur_base_config] --> Site[picknik_ur_site_config]
Base --> Mock[picknik_ur_mock_hw_config]
Site --> Gazebo[picknik_ur_gazebo_config]
Site --> Picknik1[picknik_001_ur5e_config]
Site --> Picknik2[picknik_002_ur5e_config]
```


nbbrooks:
I know this version of the PR was just to get the files online, but we should discuss what the hierarchy of base/site configs in this repo with regard to our HW configs and the tutorial examples configs. :+1:

sea-bass:
> I know this version of the PR was just to get the files online, but we should discuss what the hierarchy of base/site configs in this repo with regard to our HW configs and the tutorial examples configs. +1

I think I plan to take over this same PR and get the new hierarchy working.

src/picknik_ur_mock_hw_config/config/config.yaml
```diff
@@ -0,0 +1,27 @@
+based_on_package: "picknik_ur_base_config"
+
+hardware:
+
+  # Be sure to set your robot's IP address.
+  ip: "0.0.0.0"
+  simulated: ${MOCK_HARDWARE:-true}
```
eholum, line 7: Since "mock" is in the name we could probably just leave this as true forever.

src/picknik_ur_gazebo_config/package.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <package format="3">
   <name>picknik_ur_gazebo_config</name>
-  <version>2.4.0</version>
+  <version>2.5.0</version>
```
eholum, line 4: Bump this in the examples package, too?

src/picknik_ur_base_config/package.xml
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0"?>
 <package format="3">
   <name>picknik_ur_base_config</name>
-  <version>2.4.0</version>
+  <version>2.5.0</version>
 
-  <description>Base config package for a UR5e that can be simulated with mock hardware</description>
+  <description>Base config package for a UR5e robot</description>
```
eholum, line 6:
```suggestion
  <description>Base config package for a UR robot</description>
```

src/picknik_ur_mock_hw_config/meshes/cnc.png
eholum, line 1: I wonder if we should consider putting these in our accessories... I think this is ok for here now but my preference would be to enable LFS for these larger files, but I am also loathe to require LFS for this repo...

eholum (changes requested):
This LOOKS amazing!

I did run into a parallel planning error that I didn't expect:

image

And am curious to hear @nbbrooks opinions about the occasional "roller coaster" maneuver when moving between pick and place...

sea-bass:
@eholum The parallel error is something I ran into sporadically as well -- am talking to @sjahr about it to see what it could be.

src/picknik_ur_base_config/objectives/move_to_joint_state.xml
```diff
@@ -3,7 +3,7 @@
     <BehaviorTree ID="Move to Joint State" _description="Single-node behavior to move to a specified joint state">
         <Control ID="Sequence">
             <Action ID="GetStringFromUser" parameter_name="move_to_joint_state.waypoint_name" parameter_value="{waypoint_name}" />
-            <Action ID="MoveToJointState" controller_names="/joint_trajectory_controller /robotiq_gripper_controller" planning_group_name="manipulator" waypoint_name="{waypoint_name}" />
+            <Action ID="MoveToJointState" controller_names="/joint_trajectory_controller /robotiq_gripper_controller" planning_group_name="manipulator" waypoint_name="{waypoint_name}"/>
```
sjahr, line 6: > @eholum The parallel error is something I ran into sporadically as well -- am talking to @sjahr about it to see what it could be.

The warning is produced [here](https://github.com/PickNikRobotics/moveit_studio/blob/18f59a926de03609106e6b5e53f67d63e20e5c6c/src/moveit_studio_behavior/src/utils/mtc_utils.cpp#L86C6-L86C35). If the parallel planning port is not set, we default to not using parallel planning at all.

To fix this you need to update your objective files like this:
```suggestion
            <Action ID="MoveToJointState" controller_names="/joint_trajectory_controller /robotiq_gripper_controller" planning_group_name="manipulator" waypoint_name="{waypoint_name}" use_all_planners="true"/>
```
If you set it to false, no parallel planning is used (not a bad thing) but the warning disappears too.

sjahr: @kineticsystem You need to apply this to all objectives that use `MoveToJointState`

sea-bass: Oh, so this was a red herring! The warning happens regardless, but the errors are because planning failed! Interesting...

src/picknik_ur_mock_hw_config/meshes/cnc.png
sea-bass: Having thought about this a little more... I'd actually want those accessories to leave the core MoveIt Studio repo because we're probably eating a bunch of MBs of Docker image size having those there.

So, having each site config workspace handle its own meshes (with/without LFS) would be my preference.

src/picknik_ur_base_config/package.xml
sea-bass: Updated here and in lots of other README/package.xmls

src/picknik_ur_base_config/objectives/looping_pick_and_place_object.xml
```diff
@@ -0,0 +1,95 @@
+<?xml version="1.0"?>
```
eholum: Does this need to be here?

If so should we have a config yaml for it so that we can at least run it in mock hw without a missing config file error?

eholum (changes requested):
Still think it looks good!

Just have the one question about the looping pick and place objective... can it just live in the ur site config?

src/picknik_ur_base_config/objectives/looping_pick_and_place_object.xml
sea-bass: It does not -- site config is where it should be. Great catch.

eholum (commented):
I think this looks good, but should hold off on merging until @nbbrooks has a chance to review.

EDIT: Unless you're going to adjust joint limits here or in a follow on PR? Happy to take another look.

sea-bass:
I've adjusted the servo speed and collision check parameters for this configuration, so now the collisions I can get into are mostly recoverable.

However, because the CNC machine / table is part of the robot URDF itself, not the scene, I could only do so much before I blocked the entire arm from thinking it's in self-collision all the time.

src/picknik_ur_mock_hw_config/config/moveit/ur_servo.yaml
```diff
@@ -0,0 +1,63 @@
...
+enable_parameter_update: true
+
+## Properties of incoming commands
+command_in_type: "unitless"  # "unitless"> in the range [-1:1], as if from joystick. "speed_units"> cmds are in m/s and rad/s
+scale:
+  # Scale parameters are only used if command_in_type=="unitless"
+  linear: 2.0  # Max linear velocity. Meters per publish_period. Unit is [m/s]. Only used for Cartesian commands.
+  rotational: 3.0  # Max angular velocity. Rads per publish_period. Unit is [rad/s]. Only used for Cartesian commands.
+  # Max joint angular/linear velocity. Rads or Meters per publish period. Only used for joint commands on joint_command_in_topic.
+  joint: 2.0
```
eholum, line 11: If we drop these even more then you can still catch collisions at 100% speed.

```suggestion
  linear: 1.0  # Max linear velocity. Meters per publish_period. Unit is [m/s]. Only used for Cartesian commands.
  rotational: 2.0  # Max angular velocity. Rads per publish_period. Unit is [rad/s]. Only used for Cartesian commands.
  # Max joint angular/linear velocity. Rads or Meters per publish period. Only used for joint commands on joint_command_in_topic.
  joint: 1.0
```

eholum (commented):
Thanks for adding! I did some more playing and have another suggestion for the servo config. 

Jogging is slow, but if we think it's too slow we can jump the start up to 50%?

src/picknik_ur_mock_hw_config/config/moveit/ur_servo.yaml
sea-bass: Good call. This lets me keep the 50% scaling default.

nbbrooks:
This is looking really good

# Documentation feedback
- Can we add the mermaid diagram block to the moveit_studio_ws README? It is a great graphic

# Trial workflow feedback
- Remove URDF collision mesh from the workpiece on the left bench. Let's have it like the current mock sim where the `Pick` position closes the gripper into the collision-less block.
- Add a collision-less workpiece to the right bench (can we make it semi transparent or have Giovanni update later to a diff color?) and update the Pick Right pose so that if you commanded the arm to that waypoint and did close gripper, it would "close" on it. 
- Planning from `Pick Left` to `Pick Right` is failing for me. I think? It is because of the workpiece collision mesh?  
- Trial workflow will involve modifying 3WP PnP to do: Pick from Left -> Place in CNC -> Home -> "Grab" from CNC (no physics) -> Place on Right
  - The current 3 Waypoint PnP BT structure looks great!
  - With that in mind/in general, I suggest we rename some waypoints to be more generic: Pick Left -> Grasp Left, Pick Right -> Grasp right, Place -> Grasp CNC
- Remove "Looping pick and place" from favorites - things that dont run in this environment should not be favorited.

# Me trying to break things
- Aside from the Move from Pick Left to Pick Right (thoughts above), planning between Waypoints seems to succeed!
- I tried to make the robot move so far into a Singularity or Collision that I could not get the robot out of it (that Dave sometimes sees), while running CPU intensive stuff in the background with 100% Jog speed and could not. Sidenote: We can actually servo into the the workbench honeycomb grid here! I could see this maybe causing some confusion when people accidentally servo deep into it with a bad camera view (or conversely, impress some people) and we might want to treat the grid as a continuous sheet.
https://github.com/PickNikRobotics/moveit_studio_ws/assets/5314817/c84a59fb-b82b-4f7d-a92d-1b4f46a4c7f8
- Can we have the starting view not be the robot equivalent of a butt shot? :laughing: 
Screenshot from 2023-07-25 17-22-20

# Semi-related, but not really
- I'm not sure if this is new, but how is the Ordering of ports determined? It is inconsistent between the Editor and the BT View. In this instance, I sadly cannot see or expand to see which waypoints are being used when running `Joint Cycle` or `3 Waypoint PNP` in the BT viewer, which is annoying.
 Screenshot from 2023-07-25 17-35-42

sea-bass:
@nbbrooks Nice feedback. I got all of it done (except the UI ports thing), except as you identified, the Grasp Left -> Grasp Right planning fails very often unless you go through a nicer intermediate point.

https://github.com/PickNikRobotics/moveit_studio_ws/assets/4603398/6485410e-ced7-432a-bb90-fe559210efee

src/picknik_ur_mock_hw_config/description/picknik_ur_machine_tending.xacro
```diff
@@ -56,7 +51,7 @@
     script_filename="$(find ur_robot_driver)/resources/ros_control.urscript"
     output_recipe_filename="$(find ur_robot_driver)/resources/rtde_output_recipe.txt"
     input_recipe_filename="$(find ur_robot_driver)/resources/rtde_input_recipe.txt">
-    <origin xyz="0 0 0.8" rpy="0 0 ${pi/2}" />
+    <origin xyz="0 0 0" rpy="0 0 ${pi/2}" />
```
eholum, line 54: Lol. I think the grid is fine, it makes some sense to have the robot be the center of the universe, anyway.

eholum (commented):
Still looks great! I played with the mermaid stuff but it seems to be a GH issue. Maybe we could change the text in the boxes or something? It's kind of jarring.

README.md
```diff
@@ -10,3 +10,23 @@ For testing purposes, you can also use the [ROS 2 Control Mock Components](https
...
 This workspace offers a reasonable starting point for those users looking to develop with MoveIt Studio using custom base and site configurations.
 For more information refer to the [online documentation](https://docs.picknik.ai/en/stable/).
+
+## Universal Robots Configuration Packages
+
+This workspace contains several MoveIt Studio configuration packages for UR arms that inherit from each other for different applications.
+
+```mermaid
+graph TB
+Base[picknik_ur_base_config] --> Site[picknik_ur_site_config]
```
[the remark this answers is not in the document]
eholum: Oh man it's a safari issue! Looks fine in chrome and Firefox.

MacOS has finally let me down.

eholum (approved):
Still LGTM.

sea-bass:
> Still LGTM.

sorry had to resolve a conflict

eholum (approved):
still still lgtm.

Also my b on the changes to just have them be deleted...

PR #77, opened 2023-08-01
Simplify CNC collision to fix motion planning failures when moving between `Grasp Left` and `Grasp Right`

Abishalini:
image

Working Video - 
https://github.com/PickNikRobotics/moveit_studio_ws/assets/17363793/0b80dd86-7bcd-46dd-a175-457f342f93d6

Also, enabled using all planners through a port. This Studio PR adds the port to `SetupMTCMoveToJointState` behavior - https://github.com/PickNikRobotics/moveit_studio/pull/4198

### DIFF
--- a/src/picknik_ur_base_config/objectives/move_to_known_pose.xml
+++ b/src/picknik_ur_base_config/objectives/move_to_known_pose.xml
@@ -18,7 +18,7 @@
             <Action ID="RetrieveWaypoint" waypoint_joint_state="{target_joint_state}" waypoint_name="{waypoint_name}"/>
             <Action ID="InitializeMTCTask" task_id="move_to_known_pose" controller_names="/joint_trajectory_controller /robotiq_gripper_controller" task="{move_to_waypoint_task}"/>
             <Action ID="SetupMTCCurrentState" task="{move_to_waypoint_task}"/>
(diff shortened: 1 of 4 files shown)


sea-bass (approved):
This is beautiful, and was very much needed.

My only comment is whether we need to sweep through other Objectives in this configuration and turn on parallel planning for `SetupMTCMoveToJointState`, but that can be done in a separate PR.

Abishalini:
I can turn on parallel planning in another PR

sea-bass (approved):
Can you update the screenshot in this PR just for traceability?

eholum:
> I can turn on parallel planning for all the objectives in another PR

I think we'll have to enable it for all of them at once, right? Otherwise we end up with objectives with missing ports?

sea-bass:
> > I can turn on parallel planning for all the objectives in another PR
> 
> I think we'll have to enable it for all of them at once, right? Otherwise we end up with objectives with missing ports?

Not with this particular flag, because the utility function it calls actually has its own fallback to false (@sjahr was forward thinking about backward compatibility)

eholum:
> Not with this particular flag, because the utility function it calls actually has its own fallback to false (@sjahr was forward thinking about backward compatibility)

Ah thank you. I have asked this exact same question twice now.

PR #78, opened 2023-08-02
Added Blender source files

kineticsystem:
I've added the Blender source files to the `meshes` folder.

I've added them there so we do not need to replicate the textures. Additionally, someone can open the blender file, do some modifications, and bake the texture in the png files used by the application without the need to copy the texture around.

To test, open the two Blender files and see if can visualize the CNC machine in both, textures included.


sea-bass (commented):
Not sure how I feel about providing a few MB of source files that end users don't need to run our examples. Probably best to have this internally to not bloat the repo size.

kineticsystem:
> Not sure how I feel about providing a few MB of source files that end users don't need to run our examples. Probably best to have this internally to not bloat the repo size.

Internally where?

eholum:
Would it be preferable to put them into `picknik_accessories` even if they aren't consumed in other places? What do you guys think?

sea-bass:
This is better -- if we don't actually install those source files, it won't affect the Docker image size either

eholum:
Moved to picknik_accessories.

PR #83, opened 2023-08-09
Favorite the AprilTag Pick Objective in GZ Sim

eholum:
It works and is cool, so it should be in the favorites list.


sea-bass (approved):
Biased approval

PR #86, opened 2023-08-10
Re-add user Docker overlay example

sea-bass:
This PR reinstates the custom user Docker overlay workflow now that it works again!

To test:

* Pull this repo down
* Ensure you've set all the right variables in the `.env` file (including the version, license, and username/UID)
* From the repo root:
  * `docker compose build` -> `docker compose up`
  * `docker compose build dev` -> `docker compose up dev`

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,15 @@
+# Defaults for the MoveIt Studio docker compose agent services.
+
+# Robot configuration package to launch
+STUDIO_CONFIG_PACKAGE=picknik_ur_mock_hw_config
+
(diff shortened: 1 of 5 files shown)


eholum (changes requested):
Looks good! Just one suggestion on the comment front. Thought I'm not sure the best way to communicate restrictions on versioning... Though once we roll this into the install, config, and launch tooling this will no longer be an issue (hopefully).

eholum (approved):

PR #97, opened 2023-09-01
Set constraint's z tolerance to 3.14 in my_constraints.yaml

adlarkin:
While playing with the constraints UI editor, I noticed that it doesn't allow for values greater than or less to pi. I've updated the z tolerance on the `my_constraints.yaml` file (used by the `Constrained Pick and Place Objective`) to be `3.14` to reflect this (having a value greater than pi means this value gets truncated to pi anyways in this context).


adlarkin:
Should this be targeted to `main`, or `2.6`? :thinking:

eholum (approved):

PR #99, opened 2023-09-05
remove the manual_grasp_link offset

marioprats:
This change fixes https://github.com/PickNikRobotics/moveit_studio/issues/4543 by removing an offset of 5 degrees that existed in the xacro file.
I have noticed that other scenes have this offset too, but I'm not familiar with those scenes and whether the offset is needed for those, so fixing it in the default `picknik_ur_mock_hw_config` scene for now.

### DIFF
--- a/src/picknik_ur_base_config/description/picknik_ur.xacro
+++ b/src/picknik_ur_base_config/description/picknik_ur.xacro
@@ -97,7 +97,7 @@
   <joint name="manual_grasp_joint" type="fixed">
     <parent link="grasp_link" />
     <child link="manual_grasp_link" />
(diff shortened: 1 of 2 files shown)


sea-bass (commented):
I think this offset was added so that the coordinate frame is aligned with the camera when doing teleoperation using the arrows on the UI's camera feed.

I actually think the right solution is (unfortunately) a more involved one, where depending on context we either use `grasp_link` (as would be useful for the imarker) or `manual_grasp_link` (as would be useful for most other tasks).

marioprats:
Thanks Sebastian for the context.
I'm going to abandon this PR and move the discussion back to the issue.

PR #101, opened 2023-09-06
Add force torque sensor zeroing in configs and Re-tare objective

dyackzan:
The force torque sensors on our robots accumulate drift quickly and we wish to call the re-tare service at the start of all objectives that make use of "force thresholding" or admittance control. These changes ensure all configs comply.

Also adding an external "Re-tare" objective as a part of the picknik_ur_site_config package


src/picknik_ur_site_config/objectives/re-tare.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<root BTCPP_format="4" main_tree_to_execute="Re-tare">
+  <!-- ////////// -->
+  <BehaviorTree ID="Re-tare" _description="Zero the force torque sensors to mitigate accumulated drift error" _favorite="true" _hardcoded="false">
```
sea-bass, line 4: The name `"Re-tare"` alone seems a little lacking in descriptiveness. What about something like "Re-Zero Force/Torque Sensor" or "Re-Zero FTS"?

Same w/ the file name

src/picknik_ur_site_config/objectives/re-tare.xml
```diff
@@ -0,0 +1,11 @@
...
+    <Control ID="Sequence" name="TopLevelSequence">
+      <Decorator ID="ForceSuccess">
```
sea-bass, line 6: You should add in the comment about why we use `ForceSuccess` in this Objective as well

sea-bass (commented):
Given that this PR creates an new Behavior Tree that does the re-zeroing, in all the other Objectives you could instead refer to it using the following:

```xml
<Subtree ID="Re-tare">
```

Should save us on lines of code and copypasta

src/picknik_ur_site_config/objectives/re-tare.xml
dyackzan: Done

src/picknik_ur_site_config/objectives/re-tare.xml
dyackzan: Done, thank you for this feedback, things definitely look better and make more sense with the added description and use of the Subtree ID references.

sea-bass (changes requested):
So it seems the `/` in the Objective name messes with the URL and this doesn't work. My bad, I should have anticipated that.

image

Works if you replace it with a dash in the name:

image

sea-bass (approved):

PR #105, opened 2023-09-13
Changes to loop ConstrainedPickandPlace objective

dyackzan:
Added in a KeepRunningUntilFailure decorator at the top of the objective so it will loop.

Made the objective go from Grasp Right -> Grasp Left waypoint rather than Grasp Left -> Gripper Down -> Grasp Right:
* If we leave the Gripper Down waypoint in between, then we don't end up doing constrained planning. The straight-line path between Grasp Left and Gripper Down in the joint space also satisfies the constraint, so constrained planning is not needed.
* From trial and error, doing constrained planning from Grasp Right -> Grasp Left succeeds much more frequently then the other way around. Likely due to the joint orientations of the UR5 arm.

Added in RetryUntilSuccessful decorators in front of the MoveToWaypoint behaviors since those fail occaisionally due to collisions that pop up during path simplification.

### DIFF
--- a/src/picknik_ur_mock_hw_config/objectives/constrained_pick_place.xml
+++ b/src/picknik_ur_mock_hw_config/objectives/constrained_pick_place.xml
@@ -2,14 +2,21 @@
 <root BTCPP_format="4" main_tree_to_execute="Constrained Pick and Place">
   <!-- ////////// -->
   <BehaviorTree ID="Constrained Pick and Place" _description="Pick and place an object with constraints on the gripper's range of motion" _favorite="true">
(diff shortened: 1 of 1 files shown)


marioprats:
I tested this and looks good to me. The extra decorators make it succeed more frequently.
Still a couple of things to investigate in my opinion, independently of this PR:
1) It would be nice to have deterministic plans. This could be done by reseeding the random number generator always with the same seed. It would make failures and successes more deterministic, and always return the same path given the same inputs.
2) A bit related to the previous point: investigate why constrained planning still fails so often for an apparently simple scene like this.

marioprats (approved):

PR #108, opened 2023-09-28
Add TAMP objectives

adlarkin:


sea-bass (commented):
I think we can keep this branch here for now for testing, but let's not merge it until we spin up our dedicated config package(s) for these Task + Motion Planning examples?

We can leave this as a draft and use the branch in our submodules for testing in the meantime?

adlarkin:
Closing this because it's actually not needed

PR #109, opened 2023-09-29
Add TRAC_IK configuration

sea-bass:
This PR adds the TRAC_IK solver as the default configuration for both planning and servoing.

### DIFF
--- a/src/picknik_ur_base_config/config/config.yaml
+++ b/src/picknik_ur_base_config/config/config.yaml
@@ -139,7 +139,7 @@ moveit_params:
     path: "config/moveit/sensors_3d.yaml"
   servo_kinematics:
     package: "picknik_ur_base_config"
-    path: "config/moveit/trac_ik_kinematics_distance.yaml"
+    path: "config/moveit/trac_ik_kinematics_speed.yaml"
   joint_limits:
     package: "picknik_ur_base_config"
     path: "config/moveit/joint_limits.yaml"
--- a/src/picknik_ur_base_config/config/config.yaml
+++ b/src/picknik_ur_base_config/config/config.yaml
@@ -130,7 +130,7 @@ moveit_params:
     path: "config/moveit/stomp_planning.yaml"
   kinematics:
     package: "picknik_ur_base_config"
-    path: "config/moveit/pick_ik_kinematics.yaml"
+    path: "config/moveit/trac_ik_kinematics_distance.yaml"
   servo:
     package: "picknik_ur_base_config"
     path: "config/moveit/ur_servo.yaml"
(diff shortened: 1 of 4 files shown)


src/picknik_ur_base_config/config/moveit/trac_ik_kinematics_speed.yaml
marioprats, line 1: Do we need this config file? it doesn't seem to be referenced from the general config.

sea-bass: Unclear... I wanted someone testing this to play around with speed vs. distance modes and for us to basically converge on one vs. the other as a default.

marioprats: Ah, ok, thanks for the clarification! LGTM

marioprats (approved):

src/picknik_ur_base_config/config/config.yaml
```diff
@@ -139,7 +139,7 @@ moveit_params:
     path: "config/moveit/sensors_3d.yaml"
   servo_kinematics:
     package: "picknik_ur_base_config"
-    path: "config/moveit/kinematics.yaml"
+    path: "config/moveit/trac_ik_kinematics_distance.yaml"
```
sea-bass, line 142:
```suggestion
    path: "config/moveit/trac_ik_kinematics_speed.yaml"

src/picknik_ur_base_config/config/moveit/trac_ik_kinematics_speed.yaml
sea-bass: Actually, I just changed servo to use the "Speed" mode because burning 5ms of computation budget on a loop meant to run at a 2ms rate is... yeah, not correct :)

sea-bass: it is, but "Speed" terminates on first solution before the timeout (which on average is 0.5 ms), whereas "Distance" will eat the whole budget and keep generating solutions.

marioprats (approved):

moriarty:
@sea-bass does that rosdep key exist?

https://github.com/search?q=repo%3Aros%2Frosdistro+trac_ik&type=code

It does not look like it was released

sea-bass:
> @sea-bass does that rosdep key exist?
> 
> https://github.com/search?q=repo%3Aros%2Frosdistro+trac_ik&type=code
> 
> It does not look like it was released

No, it was added as a submodule in the Studio repo!

PR #110, opened 2023-10-04
remove trac_ik_kinematics_plugin exec_depend

moriarty:
not a valid rosdep key:
https://github.com/search?q=repo%3Aros%2Frosdistro+trac_ik&type=code


sea-bass:
The repo was added as a submodule and built from source in the latest MoveIt Studio image. No need to remove this dependency!

moriarty:
https://bitbucket.org/traclabs/trac_ik.git/src

PR #112, opened 2023-10-11
Add 'Move along Path' objective

marioprats:
This is a demo objective showing how to use `PlanCartesianPath` + `Execute Follow Joint Trajectory`.
It creates a rectangle covering two cells of the 0.4m grid in Studio and moves along the rectangle contour with a 6DOF Cartesian constraint.
Closes https://github.com/PickNikRobotics/moveit_studio/issues/4669.

move_along_path


src/picknik_ur_mock_hw_config/objectives/move_along_path.xml
```diff
@@ -0,0 +1,21 @@
...
+      <Action ID="CreateStampedPose" reference_frame="base_link" position_xyz="0.4;0.4;-0.0" orientation_xyzw="0.0;1.0;0.0;0.0" stamped_pose="{stamped_pose}"/>
+      <Action ID="AddPoseStampedToVector" input_pose="{stamped_pose}" pose_stamped_vector="{pose_stamped_vector}"/>
+      <Action ID="CreateStampedPose" reference_frame="base_link" position_xyz="0.4;0.8;0.0" orientation_xyzw="0.0;1.0;0.0;0.0" stamped_pose="{stamped_pose}"/>
+      <Action ID="AddPoseStampedToVector" input_pose="{stamped_pose}" pose_stamped_vector="{pose_stamped_vector}"/>
+      <Action ID="CreateStampedPose" reference_frame="base_link" position_xyz="-0.4;0.8;0.0" orientation_xyzw="0.0;1.0;0.0;0.0" stamped_pose="{stamped_pose}"/>
+      <Action ID="AddPoseStampedToVector" input_pose="{stamped_pose}" pose_stamped_vector="{pose_stamped_vector}"/>
+      <Action ID="CreateStampedPose" reference_frame="base_link" position_xyz="-0.4;0.4;-0.0" orientation_xyzw="0.0;1.0;0.0;0.0" stamped_pose="{stamped_pose}"/>
+      <Action ID="AddPoseStampedToVector" input_pose="{stamped_pose}" pose_stamped_vector="{pose_stamped_vector}"/>
```
sea-bass, line 16: My only suggestion is bundle this all up into a subtree (or at least a separate Sequence node) so that the entire functionality that is "path creation" is bundled into a collapsible entity.

marioprats: Nice trick, thanks!

src/picknik_ur_mock_hw_config/objectives/move_along_path.xml
```diff
@@ -0,0 +1,21 @@
...
+      <Action ID="PlanCartesianPath" path="{pose_stamped_vector}" planning_group_name="manipulator" tip_link="grasp_link" tip_offset="0.0;0.0;0.0" position_only="false" blending_radius="0.02" velocity_scale_factor="1.0" acceleration_scale_factor="1.0" trajectory_sampling_rate="100" joint_trajectory_msg="{joint_trajectory_msg}"/>
+      <Action ID="ExecuteFollowJointTrajectory" execute_follow_joint_trajectory_action_name="/joint_trajectory_controller/follow_joint_trajectory" joint_trajectory_msg="{joint_trajectory_msg}"/>
```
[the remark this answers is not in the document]
eholum: > Is there a trick to remove a key-value from the blackboard?

I am not sure about this. I don't think that there's a generic way to "unset" a blackboard variable from a behavior but someday that might be worth adding.

How were you running the loop? If I just put the `MoveToWaypoint` and the `ExecuteFollowJointTrajectory` behaviors in a loop you shouldn't have to change anything about the vector, right?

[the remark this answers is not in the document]
sea-bass: ^ what he said.

But if you did need to push new stamped poses to a vector in a loop, I think the solution is to make a new Behavior that initializes an empty vector of PoseStamped messages.

src/picknik_ur_mock_hw_config/objectives/move_along_path.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<root BTCPP_format="4" main_tree_to_execute="Move along Path">
+  <!-- ////////// -->
+  <BehaviorTree ID="Move along Path" _description="Moves the robot tip on a rectangular path." _favorite="true" _hardcoded="false">
+    <Control ID="Sequence" name="TopLevelSequence">
+      <Action ID="MoveToWaypoint" waypoint_name="Look at Right Table" planning_group_name="manipulator" controller_names="/joint_trajectory_controller" use_all_planners="false" constraints=""/>
```
MarqRazz, line 6: Should we be concerned about people changing/moving this waypoint?

src/picknik_ur_mock_hw_config/objectives/move_along_path.xml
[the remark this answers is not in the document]
marioprats: aha, thanks! it was kind of obvious :sweat_smile: 
Ok I turned the plan + execution part into a loop now.

src/picknik_ur_mock_hw_config/objectives/move_along_path.xml
marioprats: This should still work even if that waypoint changes a bit. But of course it won't work if it's a very different waypoint.
We could use `CreateJointState` + `MoveToJointState` instead if you think that's better. But it would be slightly less portable to other robots. With the current UI it is easier to visualize/create a named waypoint than it is to create a custom joint state message. Also it looks like those predefined waypoints are used in other behaviors as well? But let me know what you prefer, happy to go one way or the other.

eholum (approved):

PR #113, opened 2023-10-11
docker-compose.yaml:  Increase rest_api container timeout

zultron:
The rest_api container health check took longer than the 1m timeout to start passing on a lightly loaded 2022 ThinkPad Carbon X1 running Ubuntu, and MoveIt Studio refused to start.

Studio v. 2.4.0 worked fine; IIRC this problem started in v. 2.5.0.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -55,7 +55,7 @@ services:
     healthcheck:
       test: "curl -f http://localhost:3200/objectives"
       interval: 5s
-      timeout: 1m
+      timeout: 3m
     command: rest_api.app
 
   # Starts the robot drivers


eholum:
Thanks for the suggestion, @zultron!

This is somewhat surprising. In `2.5.0` we removing auto-building of the user's mounted workspace, so starting should have gotten _faster_ after running `./moveit_studio build`. We're also running on much less spec'd machines than the Carbon X1... maybe something else is going on. Are you running on exactly `2.5.0`? There were some issues with folder mounts in the studio containers that were fixed in `2.5.1` and `2.5.2`. I would bump minor versions, and maybe with a clean install from:

https://docs.picknik.ai/en/2.5.2/getting_started/setup_tutorials/install_software/software_installation.html#quick-start

And maybe that will speed things up for you?

sea-bass:
Closing due to inactivity. Feel free to reopen as needed!

PR #118, opened 2023-11-01
Remove Gibson behavior plugin loader

EzraBrooks:
This is going away in favor of more core-to-Studio ML behaviors.


EzraBrooks:
Later.

PR #122, opened 2023-11-07
Add comment for `moveit_params.joint_group_name` config parameter

sea-bass:
This adds a comment to the `moveit_params.joint_group_name` config parameter to explain that it's used for waypoint saving.

Also updates another doc link.


sea-bass:
Just going to go ahead and merge this.

PR #124, opened 2023-11-10
Update paths for our 2f_85 gripper URDF

eholum:
These were updated in: https://github.com/PickNikRobotics/ros2_robotiq_gripper/commit/8e4025879957a7568afd870c1063010464bd3249
, but not in our gripper URDF.

Needed for properly rendering grasp previews:

image

### DIFF
--- a/src/picknik_ur_base_config/description/robotiq_2f_85_gripper.urdf
+++ b/src/picknik_ur_base_config/description/robotiq_2f_85_gripper.urdf
@@ -229,13 +229,13 @@
     <visual>
       <origin rpy="0 0 0" xyz="0 0 0"/>
       <geometry>
-        <mesh filename="package://robotiq_description/meshes/visual/ur_to_robotiq_adapter.dae"/>
+        <mesh filename="package://robotiq_description/meshes/visual/2f_85/ur_to_robotiq_adapter.dae"/>
       </geometry>
     </visual>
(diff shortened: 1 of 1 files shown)


MarqRazz (commented):
Why are we keeping a copy of the grippers URDF here? Is there a reason we can't use the xacros included in robotiq_description ([like this example here)](https://github.com/PickNikRobotics/ros2_robotiq_gripper/blob/main/robotiq_description/urdf/robotiq_2f_85_gripper.urdf.xacro)?

eholum:
> Why are we keeping a copy of the grippers URDF here? Is there a reason we can't use the xacros included in robotiq_description ([like this example here)](https://github.com/PickNikRobotics/ros2_robotiq_gripper/blob/main/robotiq_description/urdf/robotiq_2f_85_gripper.urdf.xacro)?

There's a longer historical answer here, but the short answer is because the UI can't parse nor process the xacro macros in the gripper description package. We still require a standalone, fully parsed URDF for rendering grasp previews. I think we can push in a change to handle processing xacros now that some blocking work has been completed. Hopefully we won't need to maintain this for too much longer...

Abishalini (approved):

Abishalini:
Tested this with the gazebo config. 
Without these changes, the rest API process crashes.

PR #128, opened 2023-11-15
Use public URLs for submodules

eholum:
So that users do not require authentication to clone.


sea-bass (approved):

marioprats:
Does this needs to be replicated in `main`?

PR #131, opened 2023-11-20
Remove Legacy Parameters for Configuration Packages

eholum:
We have removed the need for various UI parameters, including setting endpoint frames, which now just comes from servo configuration in general. I have also removed servo params that have been removed since the migration to new MoveIt Servo.

### DIFF
--- a/src/picknik_ur_base_config/config/moveit/ur_servo.yaml
+++ b/src/picknik_ur_base_config/config/moveit/ur_servo.yaml
@@ -2,9 +2,6 @@
 # Modify all parameters related to servoing here
 ###############################################
 
-# Enable dynamic parameter updates
-enable_parameter_update: true
-
 ## Properties of incoming commands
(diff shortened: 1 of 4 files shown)


src/picknik_ur_base_config/config/moveit/ur_servo.yaml
```diff
@@ -32,7 +29,6 @@ publish_joint_accelerations: false
 ## Plugins for smoothing outgoing commands
 use_smoothing: true
 smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilterPlugin"
-low_pass_filter_coeff: 1.5  # Larger --> trust the filtered data more, trust the measurements less.
```
sea-bass, line 35: while it is gone from here, it can still be passed in as follows: https://github.com/ros-planning/moveit2/pull/2114/files

... unsure how we should handle this going forward, though.

sea-bass (approved):

PR #134, opened 2023-11-28
Bump Robotiq Gripper Driver Version

eholum:
We need to pull in the major refactor from @kineticsystem, this submodule bump does that.


moriarty (approved):

eholum:
Oh gosh we should definitely change the base on this. Wait one...

eholum:
@moriarty sorry for the trouble... that was out of date. This is the real big bump...

moriarty (approved):
Sorry Missed this review before signing off for holidays

PR #136, opened 2023-11-30
Replace IsForceWithinThreshold with ForceExceedsThreshold

schornakj:
The previous IsForceWithinThreshold behavior was configured using a YAML file, while the new ForceExceedsThreshold behavior which replaces it gets these values from data ports.

This PR updates use of this behavior to match the new pattern.


sea-bass (approved):
LGTM.

sea-bass (changes requested):
Sorry, approved too early.

This actually will change the force threshold parameters in Gazebo configs, since we're no longer using config files.

So what you'll need to do is actually create full copies of the `close_cabinet_door.xml`, `push_button.xml`, and `push_button_ml.xml` in the Gazebo config package, which contain the different parameters (10 N threshold / 10 consecutive values) for `ForceExceedsThreshold` nodes.

This is one downside of moving away from YAML, I guess, but I think it's a step in the right direction regardless.

PR #139, opened 2023-12-05
Scan and plan objective

uavster:
* The scan and plan config includes custom waypoints and objectives
* New Objective definition
* Trajectory file


uavster:
@eholum Retargeted this one to v2.11, too.

src/picknik_ur_gazebo_scan_and_plan_config/config/config.yaml
```diff
@@ -9,3 +9,14 @@ based_on_package: "picknik_ur_gazebo_config"
 optional_feature_params:
   gazebo_world_package_name: "picknik_ur_gazebo_scan_and_plan_config"
   gazebo_world_path: "description/simulation_worlds/scan_and_plan_world.sdf"
+
+objectives:
+  # Override with a new set of waypoints based on the Gazebo world.
+  waypoints_file:
```
uavster, line 15: @sea-bass Since this inherits from the gazebo config, it gets all Objectives there, but overrides the waypoints. I removed the waypoints that no longer made sense, given that we don't have an ISS, boxes or april tags. However, I can't do the same with the Objectives, and I guess there are Objectives that may use some of the waypoints I removed? 

Maybe not something for this PR because of the urgency, but something to keep in mind for a later cleanup. Do you agree?

sea-bass: I agree. The "right" thing to do here would be to basically have a bare-bones Gazebo config package, and then have both the space station and scan + plan configs inherit from that, each adding their own unique Objectives.

Let's press on for now.

src/picknik_ur_gazebo_scan_and_plan_config/objectives/trajectory_around_hercules.yaml
```diff
@@ -0,0 +1,41 @@
+pose2:
```
sea-bass, line 1: Can we have `pose1` before `pose2` in this file?

sea-bass (changes requested):
Things look good, but the scan and plan Objective right now is basically one giant sequence with a lot of nodes.

I would strongly recommend trying to break these down into subtrees that can be collapsed/expanded, each of which has its own functional component.

Minimally, we'd want:
* Localize object using ICP registration
* For loop over all the trajectory points:
  * MTC in its own subtree

These can be "local" subtrees; meaning they don't necessarily need to be in their own XML. But this at least helps manage complexity and give users the chance to expand/collapse sub-components more easily.

src/picknik_ur_gazebo_scan_and_plan_config/objectives/trajectory_around_hercules.yaml
uavster: Done.

sea-bass (approved):
Thanks for the subtrees! Assuming you've tested it.

uavster:
> Thanks for the subtrees! Assuming you've tested it.

Yes

src/picknik_ur_gazebo_scan_and_plan_config/objectives/trajectory_around_hercules.yaml
```diff
@@ -0,0 +1,41 @@
+pose1:
+ header:
+  frame_id: "world"
```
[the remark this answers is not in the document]
uavster: Agreed. This comes from the constraints we impose when transforming poses, and we need to improve it. Some discussion on the topic [here](https://github.com/PickNikRobotics/moveit_studio/pull/5507#issuecomment-1834852288).

src/picknik_ur_gazebo_scan_and_plan_config/objectives/object_inspection.xml
```diff
@@ -0,0 +1,47 @@
...
+      <Action ID="InitializeMTCTask" task_id="move_to_pose" controller_names="/joint_trajectory_controller" task="{move_to_pose_task}"/>
+      <Action ID="SetupMTCCurrentState" task="{move_to_pose_task}"/>
+      <Action ID="SetupMTCMoveToPose" ik_frame="manual_grasp_link" planning_group_name="manipulator" target_pose="{target_pose}" use_all_planners="false" task="{move_to_pose_task}"/>
+      <Action ID="PlanMTCTask" task="{move_to_pose_task}" solution="{move_to_pose_solution}"/>
+      <Action ID="ExecuteMTCTask" solution="{move_to_pose_solution}"/>
+    </Control>
+  </BehaviorTree>
+  <BehaviorTree ID="EstimateObjectPose">
+    <Control ID="Sequence">
+      <Action ID="LoadPointCloudFromFile" file_path="{object_model_file_path}" frame_id="world" num_sampled_points="10000" random_seed="1234" point_cloud="{model_point_cloud}" scale="1.0" color="255;150;0"/>
```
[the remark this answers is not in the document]
MarqRazz: `Failed to find match for field 'rgb'.` is a message from PCL when loading the .stl file.

PR #142, opened 2023-12-07
Remove duplicate Objective

moriarty:
This exists in the base picknik_ur_site_config

fixes https://github.com/PickNikRobotics/moveit_studio/issues/5537


sea-bass:
@chancecardona Do you mind taking a look here to determine whether this should be deleted, renamed, other?

chancecardona (approved):

PR #149, opened 2023-12-14
Initial CI Job

moriarty:
Only does a colcon build in the studio container

- rely on github setting workdir to /github/workspace and mounting the code there

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,11 +14,11 @@ jobs:
     container: picknikciuser/moveit-studio:main
     steps:
       - uses: actions/checkout@v4
-      - name: Run colcon build
+      - name: Run colcon build and test
         run: |
           . /opt/ros/humble/setup.sh
           . /opt/underlay_ws/install/setup.sh
           . /opt/overlay_ws/install/setup.sh
-          pwd
-          ls -la .
           colcon build
+          colcon test
+          colcon test-result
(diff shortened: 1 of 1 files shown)


moriarty:
I've gotten to a point where I can't test using https://github.com/nektos/act/ and will peel parts into here

moriarty:
This needs to be squash merged :see_no_evil:

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,24 @@
...
+  integration-test-in-studio-container:
+    runs-on: ubuntu-latest
+    env:
+      USER_WS: /github/workspace
+    # TODO: docker tag
+    # TODO: example for custom base images
+    container: picknikciuser/moveit-studio:main
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run colcon build and test
```
eholum, line 17: Maybe worth separating into build and test? It's nice to always get test results, for instance, and this won't do it if the tests fail I think.


```
      - name: Run tests
        run: |
          . /opt/ros/humble/setup.sh
          . /opt/overlay_ws/install/setup.sh
          colcon test
      - if: always()
        run: colcon test-result --verbose
```

moriarty: oh yes actually I have it like that in another branch... I've been peeling off changes to get this in

PR #151, opened 2023-12-16
Add trajectory preview subtree for convenience

pac48:
This PR address https://github.com/PickNikRobotics/moveit_studio/issues/3545
I added a new objective to the `picknik_ur_base_config` package to preview a trajectory if the user is available.

### DIFF
--- a/src/picknik_ur_base_config/objectives/wait_for_trajectory_approval_if_user_available.xml
+++ b/src/picknik_ur_base_config/objectives/wait_for_trajectory_approval_if_user_available.xml
@@ -5,7 +5,7 @@
       <Inverter>
         <Action ID="IsUserAvailable"/>
       </Inverter>
-      <Action ID="WaitForUserTrajectoryApproval" solution="{mtc_solution}"/>
+      <Action ID="WaitForUserTrajectoryApproval" solution="{solution}"/>
     </Fallback>
   </BehaviorTree>
 </root>
--- a/src/picknik_ur_base_config/objectives/wait_for_trajectory_approval_if_user_available.xml
+++ b/src/picknik_ur_base_config/objectives/wait_for_trajectory_approval_if_user_available.xml
@@ -5,7 +5,7 @@
       <Decorator ID="Inverter">
         <Action ID="IsUserAvailable"/>
       </Decorator>
-      <Action ID="WaitForUserTrajectoryApproval" solution="{mtc_solution}"/>
+      <Action ID="WaitForUserTrajectoryApproval" solution="{solution}"/>
     </Control>
   </BehaviorTree>
 </root>
--- a/src/picknik_ur_base_config/objectives/wait_for_trajectory_approval_if_user_available.xml
+++ b/src/picknik_ur_base_config/objectives/wait_for_trajectory_approval_if_user_available.xml
@@ -1,0 +1,11 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<root BTCPP_format="4" main_tree_to_execute="Wait for Trajectory Approval if User Available">
+  <BehaviorTree ID="Wait for Trajectory Approval if User Available" _description="Preview a shared pointer to an MTC Solution object via an input data port if a user is available.">
+    <Fallback name="wait_for_approval_if_user_available">
+      <Inverter>
+        <Action ID="IsUserAvailable"/>
+      </Inverter>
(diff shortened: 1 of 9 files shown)


src/picknik_ur_base_config/objectives/wait_for_trajectory_approval_if_user_available.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<root BTCPP_format="4" main_tree_to_execute="Wait for Trajectory Approval if User Available">
+  <BehaviorTree ID="Wait for Trajectory Approval if User Available" _description="Preview a shared pointer to an MTC Solution object via an input data port if a user is available.">
+    <Fallback name="wait_for_approval_if_user_available">
+      <Inverter>
+        <Action ID="IsUserAvailable"/>
+      </Inverter>
+      <Action ID="WaitForUserTrajectoryApproval" solution="{mtc_solution}"/>
```
sea-bass, line 8:
```suggestion
      <Action ID="WaitForUserTrajectoryApproval" solution="{solution}"/>

File: src/picknik_ur_base_config/objectives/wait_for_trajectory_approval_if_user_available.xml [hunk 2]
```diff
@@ -0,0 +1,11 @@
...
+    </Fallback>
```
sea-bass, line 9: We actually use the complete syntax here.

```suggestion
    <Control ID="Fallback" name="wait_for_approval_if_user_available">
      <Decorator ID="Inverter">
        <Action ID="IsUserAvailable"/>
      </Decorator>
      <Action ID="WaitForUserTrajectoryApproval" solution="{mtc_solution}"/>
    </Control>
```

sea-bass (changes requested):
The subtree looks great -- but I think we should also use it as a subtree in all the Objectives that currently use it!

For example, in [this objective](https://github.com/PickNikRobotics/moveit_studio_ws/blob/main/src/picknik_ur_site_config/objectives/pick_object.xml#L15-L20).

old:
```
            <Fallback name="wait_for_approval_if_user_available">
                <Inverter>
                    <Action ID="IsUserAvailable"/>
                </Inverter>
                <Action ID="WaitForUserTrajectoryApproval" solution="{pick_object_solution}"/>
            </Fallback>
            <Action ID="ExecuteMTCTask" solution="{pick_object_solution}"/>
```

new:
```
            <SubTree ID="Wait for Trajectory Approval if User Available" solution="{mtc_solution}">
```

It should end up looking something like this:

image

pac48:
> The subtree looks great -- but I think we should also use it as a subtree in all the Objectives that currently use it!
> 
> For example, in [this objective](https://github.com/PickNikRobotics/moveit_studio_ws/blob/main/src/picknik_ur_site_config/objectives/pick_object.xml#L15-L20).
> 
> old:
> 
> ```
>             <Fallback name="wait_for_approval_if_user_available">
>                 <Inverter>
>                     <Action ID="IsUserAvailable"/>
>                 </Inverter>
>                 <Action ID="WaitForUserTrajectoryApproval" solution="{pick_object_solution}"/>
>             </Fallback>
>             <Action ID="ExecuteMTCTask" solution="{pick_object_solution}"/>
> ```
> 
> new:
> 
> ```
>             <SubTree ID="Wait for Trajectory Approval if User Available" solution="{mtc_solution}">
> ```
> 
> It should end up looking something like this:
> 
> image

Should the new block be like this instead? I don't see where {mtc_solution} would be referenced from.
```            
<SubTree ID="Wait for Trajectory Approval if User Available" solution="{pick_object_task}">
```

sea-bass:
> Should the new block be like this instead? I don't see where {mtc_solution} would be referenced from.
> 
> ```
> <SubTree ID="Wait for Trajectory Approval if User Available" solution="{pick_object_task}">
> ```

Yes, good call, but it should be `"{pick_object_solution}"` based on what I see.

pac48:
> > Should the new block be like this instead? I don't see where {mtc_solution} would be referenced from.
> > ```
> > <SubTree ID="Wait for Trajectory Approval if User Available" solution="{pick_object_task}">
> > ```
> 
> Yes, good call, but it should be `"{pick_object_solution}"` based on what I see.

Great, I added `"pick_object_solution"` and fixed the rest.

adlarkin (approved):

PR #154, opened 2023-12-20
Change Objectives due to reworked UI parameters

sea-bass:
This replaces the `GetPoseFromUser` / `GetJointStateFromUser` Behaviors in Objectives that get called from the UI.

Also, `Move to Known Pose` has been replaced with `Move to Joint State` since the UI can do the conversion from waypoint name to joint state without having to involve the backend.


dyackzan (approved):
Tested and confirmed this works with parameter manager changes in https://github.com/PickNikRobotics/moveit_studio/pull/5702

PR #156, opened 2023-12-20
Use CYCLONE DDS else bad things happen

moriarty:
Using the Default DDS Vendor Fast DDS causes things to segfault :( Also source the current colcon workspace install space for python path

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,11 +11,30 @@ jobs:
       # TODO this variable is required but github workspace is different
       # see https://github.com/actions/runner/issues/2058
       USER_WS: /github/workspace
+      ROS_DOMAIN_ID: 49
+      ROS_LOCALHOST_ONLY: 1
+      RMW_IMPLEMENTATION: rmw_cyclonedds_cpp
+      CYCLONEDDS_URI: /tmp/cyclonedds.xml
     # TODO: docker tag
     # TODO: example for custom base images
     container: picknikciuser/moveit-studio:main
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -11,11 +11,30 @@ jobs:
...
     # TODO: docker tag
     # TODO: example for custom base images
     container: picknikciuser/moveit-studio:main
     steps:
       - uses: actions/checkout@v4
+      # The lo interfaces will not have multicast, and will barf trying to claim a participant index
+      # with the default cyclone dds config. We set one for the unit tests to be able to execute.
+      - name: Configure DDS
+        run: |
+          cat <<EOF > /tmp/cyclonedds.xml
```
sea-bass, line 27: Can this be an actual file that's part of this repo or the upstream repo, instead of a long multiline string in this YAML?

moriarty: No because it's suppose to be easy to let end users override this with their own

sea-bass: Right, so wouldn't it be easier to show an end user how to include a discrete DDS config file in their repo that gets copied into this target location for CI, or is already set up in their overlay image?

moriarty: I believe that's already setup but here we're explicitly setting CYCLONEDDS_URI to this temp file, and then writing the temp file here... just for CI

sea-bass: I see... well, let's just keep it as is, then!

sea-bass (approved):

PR #160, opened 2023-12-22
Add cohesive Teleoperate Objective

sea-bass:
This reworks teleoperation by using a single Objective that allows switching modes between the UI, vs. a bunch of disparate Objectives for each time of user control.

Critically, this allows teleoperation not only as a standalone UI interaction with the "Teleoperate Standalone" Objective, but as part of an Objective by bringing in "Teleoperate" as a subtree.

I'm not super married to the naming, and very open to suggestions. Working changes now are "Teleoperate" (top-level) and "Request Teleoperation" (subtree)


src/picknik_ur_base_config/config/config.yaml
```diff
@@ -126,6 +126,9 @@ moveit_params:
   ompl_planning:
     package: "picknik_ur_base_config"
     path: "config/moveit/ompl_planning.yaml"
+  pilz_planning:
+    package: "picknik_ur_base_config"
+    path: "config/moveit/pilz_industrial_motion_planner_planning.yaml"
```
sea-bass, line 131:
```suggestion
```

sea-bass:
NOTE: Integration tests failing here is expected because we're calling a new Behavior that's part of an upstream PR for this. I think it might be beneficial for me to maybe break that out and merge it first...

PR #161, opened 2023-12-22
Add pilz planning config

sjahr:


sea-bass (approved):

Abishalini:
We need to get this into Studio main so that objectives like `Move to Known Pose` work

PR #165, opened 2024-01-05
Update README to show scan and plan config

sea-bass:
This PR adds information about the new scan and plan configuration to the README.


uavster (approved):
Thanks for this update! LGTM. The failed integration test seems unrelated.

PR #166, opened 2024-01-05
Fix integration tests to use new parameter storage

sea-bass:


src/picknik_ur_studio_integration_testing/test/test_basic_ur5e_objectives.cpp
```diff
@@ -170,24 +140,53 @@ TEST_F(ObjectiveFixture, DISABLED_TestInspectSurface)
   EXPECT_TRUE(sendDoObjectiveSequenceGoal(std::move(do_objective_goal), 30.0));
 }
 
+TEST_F(ObjectiveFixture, TestMoveToJointState)
```
sea-bass, line 143: should I also add a `TestMoveToPose` test?

Abishalini (approved):

PR #169, opened 2024-01-08
Update behavior names for Open Door objectives

Abishalini:
A bunch of behaviors were renamed in this PR - https://github.com/PickNikRobotics/moveit_studio/pull/5282
I'm not sure why this was not caught while testing v2.11


chancecardona (approved):
Nice, good catch. These behaviors will be going away soon for their updated versions here: https://github.com/PickNikRobotics/moveit_studio_ws/pull/132 but in the mean time this change lets the Gibson behaviors work again so LGTM.

PR #171, opened 2024-01-09
Remove Gibson and Visual Servo. 

chancecardona:
This is part of https://github.com/PickNikRobotics/moveit_studio/pull/5843 and removes these two no longer used dependencies. I'm not sure if this is correctly how you define an empty plugin list in yaml though so worth double checking.


eholum (approved):

eholum (changes requested):
Actually can we delete the servoing objective?

src/picknik_ur_site_config/config/config.yaml
```diff
@@ -6,11 +6,7 @@ hardware:
   ip: "0.0.0.0"
 
 objectives:
-  behavior_loader_plugins:
-    visual_servo_behaviors:
-      - "moveit_visual_servo::behaviors::VisualServoBehaviorLoader"
-    gibson:
-      - "gibson_behavior::GibsonBehaviorsLoader"
+  behavior_loader_plugins: []
```
EzraBrooks, line 9: I think this is of type dictionary, not array.

```suggestion
  behavior_loader_plugins: {}
```

sea-bass (changes requested):
The integration test fixes are in https://github.com/PickNikRobotics/moveit_studio_ws/pull/166, so no need to do anything here -- instead we need to merge down `v3.0` into `main`. So revert those changes here and we can merge this in.

src/picknik_ur_studio_integration_testing/picknik_ur_studio_integration_testing/generate_agent_plus_drivers_launch_description.py
```diff
@@ -81,7 +81,7 @@ def generate_agent_plus_drivers_launch_description(
 
     # Get path to test objective directory defined within this repo
     test_objective_path = Path(
-        get_package_share_directory("moveit_studio_integration_testing"),
+        get_package_share_directory("picknik_ur_studio_integration_testing"),
```
sea-bass, line 84: revert this change as it's incoming in other PRs

src/picknik_ur_studio_integration_testing/picknik_ur_studio_integration_testing/generate_agent_plus_drivers_launch_description.py
```diff
@@ -110,7 +110,7 @@ def generate_agent_plus_drivers_launch_description(
...
```
sea-bass, line 113: same here, revert this change as it's incoming in other PRs

sea-bass (approved):

PR #173, opened 2024-01-15
Update servo parameters - studio_ws

pac48:
This PR allows Studio to use the new acceleration-based smoothing plugin for servo. The hard joint limits need to be tested and verified.

Requires:
- [x] https://github.com/ros-planning/moveit2/pull/2651

### DIFF
--- a/src/picknik_ur_base_config/config/control/picknik_ur.ros2_control.yaml
+++ b/src/picknik_ur_base_config/config/control/picknik_ur.ros2_control.yaml
@@ -14,7 +14,7 @@ controller_manager:
     robotiq_activation_controller:
       type: robotiq_controllers/RobotiqActivationController
     streaming_controller:
-      type: position_controllers/JointGroupPositionController
+      type: joint_trajectory_controller/JointTrajectoryController
     admittance_controller_open_door:
       type: admittance_controller/AdmittanceController
     joint_trajectory_controller_chained_open_door:
@@ -131,6 +131,32 @@ streaming_controller:
       - position
     state_interfaces:
       - position
+      - velocity
+    command_joints:
+      - shoulder_pan_joint
+      - shoulder_lift_joint
+      - elbow_joint
+      - wrist_1_joint
+      - wrist_2_joint
+      - wrist_3_joint
+    state_publish_rate: 100.0
+    action_monitor_rate: 20.0
+    allow_partial_joints_goal: false
+    constraints:
+      stopped_velocity_tolerance: 0.0
+      goal_time: 0.0
+      shoulder_pan_joint:
+        goal: 0.05
+      shoulder_lift_joint:
+        goal: 0.05
+      elbow_joint:
+        goal: 0.05
+      wrist_1_joint:
+        goal: 0.05
+      wrist_2_joint:
+        goal: 0.05
+      wrist_3_joint:
+        goal: 0.05
 
 robotiq_gripper_controller:
   ros__parameters:
--- a/src/picknik_ur_base_config/config/control/picknik_ur.ros2_control.yaml
+++ b/src/picknik_ur_base_config/config/control/picknik_ur.ros2_control.yaml
@@ -13,7 +13,7 @@ controller_manager:
       type: position_controllers/GripperActionController
     robotiq_activation_controller:
       type: robotiq_controllers/RobotiqActivationController
-    streaming_controller:
+    servo_controller:
       type: joint_trajectory_controller/JointTrajectoryController
     admittance_controller_open_door:
       type: admittance_controller/AdmittanceController
@@ -118,7 +118,7 @@ joint_trajectory_controller_chained_open_door:
       wrist_3_joint:
         goal: 0.25
 
-streaming_controller:
+servo_controller:
   ros__parameters:
     joints:
       - shoulder_pan_joint
--- a/src/picknik_ur_base_config/config/moveit/hard_joint_limits.yaml
+++ b/src/picknik_ur_base_config/config/moveit/hard_joint_limits.yaml
@@ -1,0 +1,87 @@
+# Joints limits
+#
+# Sources:
+#
+#  - Universal Robots e-Series, User Manual, UR5e, Version 5.8
+#    https://s3-eu-west-1.amazonaws.com/ur-support-site/69091/99404_UR5e_User_Manual_en_Global.pdf
+#  - Support > Articles > UR articles > Max. joint torques
+#    https://www.universal-robots.com/articles/ur-articles/max-joint-torques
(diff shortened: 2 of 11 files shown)


src/picknik_ur_base_config/config/control/picknik_ur.ros2_control.yaml
```diff
@@ -14,7 +14,7 @@ controller_manager:
     robotiq_activation_controller:
       type: robotiq_controllers/RobotiqActivationController
     streaming_controller:
-      type: position_controllers/JointGroupPositionController
+      type: joint_trajectory_controller/JointTrajectoryController
```
sjahr, line 17: I think calling a JTC-type controller a streaming controller is confusing. If we need this for servo, you should rename the controller completely. Can't you use `joint_trajectory_controller` or does it one have a specific config?

src/picknik_ur_base_config/config/control/picknik_ur.ros2_control.yaml
```diff
@@ -131,6 +131,32 @@ streaming_controller:
       - position
     state_interfaces:
       - position
+      - velocity
```
sjahr, line 134: Looks like this config is a similar to `joint_trajectory_controller`. Can't you use it instead? Is the streaming controller only used by servo? In that case we could remove it completely :thinking:

src/picknik_ur_base_config/config/moveit/hard_joint_limits.yaml
```diff
@@ -0,0 +1,87 @@
...
+    max_velocity: !degrees  30.0
+    min_position: !degrees -270.0
+  elbow_joint:
+    has_acceleration_limits: true
+    has_effort_limits: true
+    has_position_limits: true
+    has_velocity_limits: true
+    max_acceleration: !degrees 90.0
+    max_effort: 150.0
+    # we artificially limit this joint to half its actual joint position limit
```
sjahr, line 40: Really nice comments with additional explanation!

src/picknik_ur_base_config/config/moveit/ur_servo.yaml
```diff
@@ -18,24 +18,26 @@ scale:
...
 publish_joint_positions: true
-publish_joint_velocities: false
+publish_joint_velocities: true
 publish_joint_accelerations: false
 
 ## Plugins for smoothing outgoing commands
 use_smoothing: true
-smoothing_filter_plugin_name: "online_signal_smoothing::ButterworthFilterPlugin"
-low_pass_filter_coeff: 1.5  # Larger --> trust the filtered data more, trust the measurements less.
+smoothing_filter_plugin_name: "online_signal_smoothing::AccelerationLimitedPlugin"
```
sjahr, line 35: Can you add a small comment in the PR description that this PR depends on the moveit2 PR that adds this and requires an dependency update of moveit2 in studio

src/picknik_ur_gazebo_config/config/control/picknik_ur5e.ros2_control.yaml
```diff
@@ -8,7 +8,7 @@ controller_manager:
     robotiq_gripper_controller:
       type: position_controllers/GripperActionController
     streaming_controller:
-      type: position_controllers/JointGroupPositionController
+      type: joint_trajectory_controller/JointTrajectoryController
```
sjahr, line 11: Same comments as above

src/picknik_ur_gazebo_config/config/moveit/ur_servo.yaml
```diff
@@ -51,7 +53,7 @@ cartesian_command_in_topic: ~/delta_twist_cmds  # Topic for incoming Cartesian t
 joint_command_in_topic: ~/delta_joint_cmds # Topic for incoming joint angle commands
 joint_topic: /joint_states
 status_topic: ~/status # Publish status to this topic
-command_out_topic: /robot_controllers/commands # Publish outgoing commands here
+command_out_topic: /streaming_controller/joint_trajectory # Publish outgoing commands here
```
sjahr, line 56: Might need an update

sjahr (changes requested):
Not sure if we need a streaming controller at all (more in the comments) but maybe I am missing something. CI probably fails due to https://github.com/ros-planning/moveit2/pull/2651 not yet being merged

pac48:
> Not sure if we need a streaming controller at all (more in the comments) but maybe I am missing something. CI probably fails due to [ros-planning/moveit2#2651](https://github.com/ros-planning/moveit2/pull/2651) not yet being merged

The problem is that the streaming controller is referred to in the behaviors, like [here](https://github.com/PickNikRobotics/moveit_studio_ws/blob/18930844cf1dbe292a8bda17ed22d1e05d2c6a5f/src/picknik_ur_base_config/objectives/teleoperate_joint_jog.xml#L5). If we want the servo controller to be configurable, then it needs to be done this way.

sjahr:
> > Not sure if we need a streaming controller at all (more in the comments) but maybe I am missing something. CI probably fails due to [ros-planning/moveit2#2651](https://github.com/ros-planning/moveit2/pull/2651) not yet being merged
> 
> The problem is that the streaming controller is referred to in the behaviors, like [here](https://github.com/PickNikRobotics/moveit_studio_ws/blob/18930844cf1dbe292a8bda17ed22d1e05d2c6a5f/src/picknik_ur_base_config/objectives/teleoperate_joint_jog.xml#L5). If we want the servo controller to be configurable, then it needs to be done this way.

Let's call it `servo_controller` in that case and add a comment that describes why it is necessary to have it in addition to the default JTC

src/picknik_ur_gazebo_config/config/moveit/ur_servo.yaml
pac48: Fixed, thanks!

pac48:
> > > Not sure if we need a streaming controller at all (more in the comments) but maybe I am missing something. CI probably fails due to [ros-planning/moveit2#2651](https://github.com/ros-planning/moveit2/pull/2651) not yet being merged
> > 
> > 
> > The problem is that the streaming controller is referred to in the behaviors, like [here](https://github.com/PickNikRobotics/moveit_studio_ws/blob/18930844cf1dbe292a8bda17ed22d1e05d2c6a5f/src/picknik_ur_base_config/objectives/teleoperate_joint_jog.xml#L5). If we want the servo controller to be configurable, then it needs to be done this way.
> 
> What do you think about calling it `servo_controller` in that case and add a comment that describes why it is necessary to have it in addition to the default JTC

@sjahr I renamed the controller to `servo_controller`. I think it is a better and more descriptive name.

sjahr (approved):
Thanks!

PR #174, opened 2024-01-16
Read .env to get studio tag

moriarty:
Read .env file to get studio tag

- Default main branch to main tag of studio

NOTE:

This change assumes that as part of the release process the .env file
is updated when releases are created to point to the correct studio tag

### DIFF
--- a/.env
+++ b/.env
@@ -5,7 +5,7 @@ STUDIO_CONFIG_PACKAGE=picknik_ur_mock_hw_config
 
 # MoveIt Studio version tag
 # Be careful! Changing this without understanding the consequences may break your installation. 
-STUDIO_DOCKER_TAG=2.11.0
+STUDIO_DOCKER_TAG=main
 
 # Licensing -- enter your MoveIt Studio license below
 STUDIO_LICENSE_KEY=
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,7 +5,17 @@ on:
   pull_request:
   workflow_dispatch:
 jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      image_tag: ${{ steps.dotenv.outputs.studio_docker_tag }}
+    steps:
+      - uses: actions/checkout@v4
+      - id: dotenv
+        uses: falti/dotenv-action@v1.0.4
+
   integration-test-in-studio-container:
+    needs: setup
     runs-on: ubuntu-latest
     env:
       STUDIO_LICENSE_KEY: ${{ secrets.STUDIO_CI_LICENSE_KEY }}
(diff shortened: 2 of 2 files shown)


moriarty:
https://github.com/PickNikRobotics/moveit_studio_ws/actions/runs/7535821548/job/20512361112?pr=174

reading the .env file and setting the container version is working becuase the .env file is wrong :laughing: 
```
Starting%20job%20container%20/usr/bin/docker%20pull%20picknikciuser/moveit-studio:2.11.0%202.11.0:%20Pulling%20from%20picknikciuser/moveit-studio%207ae75f326557:%20Pulling%20fs%20layer
```

.env
```diff
@@ -5,7 +5,7 @@ STUDIO_CONFIG_PACKAGE=picknik_ur_mock_hw_config
 
 # MoveIt Studio version tag
 # Be careful! Changing this without understanding the consequences may break your installation. 
-STUDIO_DOCKER_TAG=2.11.0
+STUDIO_DOCKER_TAG=main
```
sea-bass, line 8: So... this `.env` file is actually used for https://docs.picknik.ai/en/stable/getting_started/configuration_tutorials/docker_configuration/create_overlay_docker_image.html, and should therefore mostly be used with a release tag here.

Is the idea that on the `main` branch this will always be `main`, but in each of the release-specific branches, like `v3.0` we would use `3.0.0` there?

.github/workflows/ci.yaml
```diff
@@ -4,8 +4,22 @@ on:
     branches: [main]
   pull_request:
   workflow_dispatch:
+  # Run every 6 hours Mon-Fri
+  schedule:
+    - chron: "* */6 * * 1-5"
```
sea-bass, line 9: when even your CI gets a weekend break 😢

moriarty: :laughing: yep CI costs $$ electricity somewhere and gotta give it a break :wink: but maybe this will be abused and more breaking changes will go in on the weekend :upside_down_face:

.env
moriarty: Yes. The main branches will always be main, but release specific branches this should point to the version that went through the release process and was tested...

This way if you back port something to a released branch the `.env` file on that release branch should have the tag set correctly and it will test against the correct version of studio.

moriarty: The tutorials get the value from [here](https://github.com/PickNikRobotics/moveit_studio/blob/main/docs_public/source/getting_started/configuration_tutorials/docker_configuration/_include/overlay_docker_env.rst) not this file specifically so if someone is following along and copy-pasting from the tutorials they should be fine.

sea-bass (approved):

.env
sea-bass: Thanks for the answer, and good point on the tutorial source being more likely to be copypasted than someone trying it out directly from here. LGTM.

PR #176, opened 2024-01-18
Small wording changes for the product update

eholum:
Avoiding changing parameter names, arguments, etc. in the overlay workflow for now. The package updates are the critical functional pieces here.


sea-bass:
I renamed the integration testing package name to not say "studio", so maybe the PR name/description should change if we keep this around.

Also, there was a PR to `picknik_accessories` that landed recently so we'll just have to deconflict those branches and ascend the submodule ladder again.

PR #178, opened 2024-01-19
Remove `manual_grasp_link`

sea-bass:
Requires https://github.com/PickNikRobotics/picknik_accessories/pull/4

### DIFF
--- a/src/picknik_ur_base_config/description/picknik_ur.xacro
+++ b/src/picknik_ur_base_config/description/picknik_ur.xacro
@@ -98,14 +98,6 @@
     <origin xyz="0.0 0.0 0.134" rpy="0.0 0.0 ${pi}" />
   </joint>
 
-  <!-- Moveit Studio requires a link name called manual_grasp_link to perform quick tasks such as inspect surface -->
-  <link name="manual_grasp_link" />
-  <joint name="manual_grasp_joint" type="fixed">
-    <parent link="grasp_link" />
-    <child link="manual_grasp_link" />
-    <origin xyz="0 0 0" rpy="${pi / 180.0 * 5} 0 0" />
-  </joint>
-
   <!-- Environment -->
   <link name="environment">
     <visual>
--- a/src/picknik_ur_base_config/description/picknik_ur.xacro
+++ b/src/picknik_ur_base_config/description/picknik_ur.xacro
@@ -91,6 +91,7 @@
     <origin xyz="0 0 0" rpy="0 0 0" />
   </xacro:robotiq_gripper>
 
+  <!-- This link aligns the axes of the grasp frame to our required convention for motion planning tasks. -->
   <link name="grasp_link" />
   <joint name="grasp_link_joint" type="fixed">
     <parent link="gripper_mount_link" />
--- a/src/picknik_ur_base_config/config/config.yaml
+++ b/src/picknik_ur_base_config/config/config.yaml
@@ -166,10 +166,9 @@ moveit_params:
 # Additional configurable parameters for the MoveIt Studio user interface.
 # [Required]
 ui_params:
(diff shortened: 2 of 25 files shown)


src/picknik_ur_base_config/description/picknik_ur.xacro
```diff
@@ -98,14 +98,6 @@
     <origin xyz="0.0 0.0 0.134" rpy="0.0 0.0 ${pi}" />
   </joint>
 
-  <!-- Moveit Studio requires a link name called manual_grasp_link to perform quick tasks such as inspect surface -->
```
MarqRazz, line 101: Should we preserve this comment for `grasp_link` in all configs?

sea-bass: The comment is still in the `gripper_and_camera.xacro`, `robotiq_2f_85_gripper.urdf`, and `config.yaml`, so I'm OK with removing this whole block.

To me, it doesn't make much sense to have a standalone comment where there no longer is a `manual_grasp_link` or `grasp_link` definition.

What do you think?

MarqRazz: Grasp link is defined on line 94 and this URDF is calling
`<xacro:include filename="$(find robotiq_description)/urdf/robotiq_2f_85_macro.urdf.xacro" />` not the one in picknik_accessories. FYI if you want to use PickNik's xacro we need the ability to pass in the `comm_port` value.

sea-bass: Gotcha! Added the comment to line 94.

MarqRazz (commented):
Tested several objectives on the `picknik_ur_gazebo_config` and they all work except for `Open Cabinet Door`. It fails with a BT error
```
Error: Failed to create behavior tree for Objective `Open Cabinet Door`. Reason: BehaviorTree error while creating tree: Possible typo? In the XML, you tried to remap port "screw_origin_pose" in node [GetMoveAlongArcSubframes / GetMoveAlongArcSubframes], but the manifest of this node does not contain a port with this name.
```

Is this something we want to fix here?

src/picknik_ur_base_config/description/picknik_ur.xacro
MarqRazz: Thanks for making the change there, but none of the robots here use that xacro which seems a little weird.

sea-bass:
I think this is failure is unrelated, and due to the fact that the ports in `GetMoveAlongArcSubframes` have been renamed without having updated this Objective in the Gazebo config.

```yaml
            <input_port name="target_grasp_pose" default="{target_grasp_pose}">Pose representing the grasp point.</input_port>
            <input_port name="hinge_axis_pose_start" default="{hinge_axis_pose_start}">Pose marking the start of the hinge axis.</input_port>
            <input_port name="hinge_axis_pose_end" default="{hinge_axis_pose_end}">Pose marking the end of the hinge axis.</input_port>
            <output_port name="move_along_arc_subframes" default="{move_along_arc_subframes}">Subframes representing the ordered screw (hinge) origin and axis, along with a grasp subframe.</output_port>
```

@chancecardona Just a heads-up, in case this is something not on your radar with your upcoming refactors (but I'm sure you've handled this).

MarqRazz (approved):

chancecardona:
> I think this is failure is unrelated, and due to the fact that the ports in GetMoveAlongArcSubframes have been renamed without having updated this Objective in the Gazebo config.

https://github.com/PickNikRobotics/moveit_studio_ur_ws/blob/184821d982e6b80bb826f3635c5c6d6fcc5cb46a/src/picknik_ur_site_config/objectives/open_cabinet_door.xml I have updated the open_cabinet_door objective now to work with the changes like you mentioned (at least until it fails to plan in sim!). So this should be taken care of by the gibson removal PR's! :+1:

PR #179, opened 2024-01-19
Add April tag support to picknik_ur_site_config

MarqRazz:
This replaces https://github.com/PickNikRobotics/picknik_ur5e_configs/pull/31 so that April tag support works on all in office UR site_configs.

Any suggestions on what I should do with the [tree_nodes_model](https://github.com/PickNikRobotics/picknik_ur5e_configs/pull/31/files#diff-b2748b68a81d9bdbcd74194e61d83a6c0d07594a9274f461baa244e9b0a52bc4) file that was added in PR#31?

### DIFF
--- a/src/picknik_ur_site_config/objectives/51mm_apriltag_detection_config.yaml
+++ b/src/picknik_ur_site_config/objectives/51mm_apriltag_detection_config.yaml
@@ -1,0 +1,9 @@
+DetectApriltags:
+  apriltag_family_name: 36h11
+  apriltag_size: 0.051
+  max_hamming: 0
+  n_threads: 1
+  quad_decimate: 1
+  quad_sigma: 0.1
+  refine_edges: false
+  z_up: true
--- a/src/picknik_ur_site_config/objectives/apriltag1_grasp_offset.yaml
+++ b/src/picknik_ur_site_config/objectives/apriltag1_grasp_offset.yaml
@@ -1,0 +1,10 @@
+GraspOffset:
+  position:
+    x: -0.1
+    y: 0.0
+    z: 0.05
+  orientation:
+    x: 0.707
+    y: -0.707
+    z: 0.0
+    w: 0.0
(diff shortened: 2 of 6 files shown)


sea-bass:
> Any suggestions on what I should do with the [tree_nodes_model](https://github.com/PickNikRobotics/picknik_ur5e_configs/pull/31/files#diff-b2748b68a81d9bdbcd74194e61d83a6c0d07594a9274f461baa244e9b0a52bc4) file that was added in PR#31?

Do we need it? I forget how this all works, but is that tree_nodes_model what ends up displaying the remapped ports in the UI when you drag in that subtree?

MarqRazz:
Tested on hardware and everything works as expected. You still need to be fairly close to the tag to get a good pose estimation (I would guess the tag was around 0.4m from the camera). You should use the `Loop Detect AprilTag` Objective to double check you get a stable estimation that aligns with the point cloud using Rviz if you're having issues.

Here is a screen shot showing the `Teach Grasp Offset From Apriltag` I ran
image

And here is the object rotated some in the scene and being picked.
[pick_april_tag.webm](https://github.com/PickNikRobotics/moveit_studio_ur_ws/assets/25058794/1198545d-f631-4dcd-a349-cdb01dc3765e)

FYI I had issues taking a snapshot before running the Objective and trajectory execution failing so I would guess the `Pick Object` Objective or included Behaviors need some love.

MarqRazz:
> Do we need it? I forget how this all works, but is that tree_nodes_model what ends up displaying the remapped ports in the UI when you drag in that subtree?

I don't think this is required. I didn't add it and the Objectives render as expected.
image

image

sea-bass:
Cool! Thanks for sharing the results here.

Want to get rid of the `tree_nodes_model.xml` if it isn't needed, then we merge this?

MarqRazz:
> Want to get rid of the tree_nodes_model.xml if it isn't needed, then we merge this?

I didn't add one in this PR and there is currently not one in the `Objectives` folder for this robot config. Is there a specific one you want me to try and remove?

sea-bass:
> > Want to get rid of the tree_nodes_model.xml if it isn't needed, then we merge this?
> 
> I didn't add one in this PR and there is currently not one in the `Objectives` folder for this robot config. Is there a specific one you want me to try and remove?

Ah, my bad, never mind then!

PR #181, opened 2024-01-22
Fix 3D object registration Objectives

sea-bass:
This PR fixes the 3D object registration Objectives in a few ways:

* Fixes the formatting of the trajectory YAML due to changes in the underlying Behaviors using the message structure
* Separates Registration and Scanning into two separate Behavior Tree XMLs
* Replaces a `MoveToPose` subtree with the simpler, single-Behavior alternative
* Updates the coordinate frame from `manual_grasp_link` to `grasp_link`, which warranted some pose changes.


uavster (approved):
Awesome!

PR #184, opened 2024-01-24
Remove pinch link tests

sea-bass:
Let's just remove these, as they're breaking CI with the latest submodule schism ... we don't want to depend on other config packages anyway.

We can re-add these in the Git repo that does have this package (which entails adding CI to it as well)


sea-bass:
The remaining integration test failure here is because we have to merge a new Behavior in an upstream PR. This one is good to go.

moriarty (approved):

PR #185, opened 2024-01-25
Fix Gazebo package dependencies

sea-bass:
Turns out the `ros_gz` package was only tagged as an `exec_depends` and not `depends`, so `rosdep install` was missing this package. Cool and good.


moriarty (approved):

eholum (commented):
it done work.

eholum:
Oh lol, I'm too slow on the trigger, apparently!

PR #189, opened 2024-01-26
Re-add streaming controller for Gazebo configuration

sea-bass:
This makes it so the Gazebo robot arm actually moves to its starting position at startup.

Somehow the UR streaming controller being active at startup did the trick...


pac48 (approved):
LGTM

PR #199, opened 2024-02-02
add visual servo Objective to ur_site_configs

marioprats:


src/picknik_ur_site_config/objectives/visual_servo_to_reference.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<root BTCPP_format="4" main_tree_to_execute="Visual Servo To Reference">
+  <!-- ////////// -->
+  <BehaviorTree ID="Visual Servo To Reference" _description="Move in closed loop to a pose relative to an AprilTag" _favorite="true" _hardcoded="false">
+    <Control ID="Sequence">
+      <Action ID="LoadObjectiveParameters" config_file_name="visual_servo_reference.yaml" parameters="{servo_reference_parameters}"/>
+      <Action ID="ActivateControllers" controller_names="servo_controller"/>
+      <Control ID="Sequence">
+        <SubTree ID="Sample April Tag" _collapsed="true" num_samples="1" tag_id="5" apriltag_config="51mm_apriltag_detection_config.yaml" max_distance="0.02" max_rotation="0.2" avg_pose="{tag_pose}"/>
```
pac48, line 9: the `tag_id` is 5 in this objective but 2 in the other one. Maybe they should both be the same.

pac48:
I tested it on the hardware and it worked great! I left one comment, but otherwise looks good.

pac48 (commented):
I tested it on the hardware and it worked great! I left one comment, but otherwise looks good.

src/picknik_ur_site_config/objectives/visual_servo_to_reference.xml
marioprats: Yeah, I had it 5 here because that's the ID of the tag I tested with.
Changed to 2 to match the gz config.

pac48 (approved):

PR #201, opened 2024-02-02
Remove hack to get around double-counted serial package installs

sea-bass:
Just found this gem which we can now remove! 🎉


moriarty (approved):
:see_no_evil:

PR #206, opened 2024-02-09
debian package install main docker-compose.yaml in /opt/moveit_pro/

orensbruli:


orensbruli:
Talking with @sea-bass I realized that it's to really needed as this is only used on CI and it currently works as is. Closing.

PR #208, opened 2024-02-14
Fix grasp link parent frames and add to gripper SRDF group

sea-bass:
This PR ensures that the gripper URDF subset and planning group is self-contained for the UI.


EzraBrooks (approved):
WFM

PR #209, opened 2024-02-15
Top level launchfile

tylerjw:
fixes https://github.com/PickNikRobotics/moveit_studio/issues/6424


Abishalini:
Should this be added to core Pro? So that this launch file can be used with all configs and not just UR mock hw

tylerjw:
I don't think it should be in the studio repo (even though it obviously could be) as this is meant to be an entry-point for users to modify and add their launch details to this launchfile.

Abishalini:
Shoot I forgot we are trying to remove launch files out of core studio. 
Can this be moved to `picknik_ur_base_config` so all the configs that inherit from it can use this launch file?

sea-bass (approved):

sea-bass (changes requested):
Sorry, I just saw @Abishalini 's comment and agree with it.

tylerjw:
This is your playground now, do whatever you want.

sea-bass (approved):

PR #219, opened 2024-02-21
Update camera sync behavior names

sea-bass:


EzraBrooks (approved):
works for me erryday.

PR #221, opened 2024-02-22
Fix Behavior loaders

sea-bass:
Since we split up the core behaviors plugins, we need to actually load them all in config!


pac48 (approved):
Thanks for doing this!

PR #223, opened 2024-02-23
Reduce UR descriptions duplication

pac48:
This PR reflects the UR site config xacros to reduce duplication https://github.com/PickNikRobotics/moveit_studio/issues/5562. This PR depends on 
- [x] https://github.com/PickNikRobotics/picknik_accessories/pull/7
- [x] https://github.com/PickNikRobotics/ros2_robotiq_gripper/pull/47

I refactor most of the  UR robots into the `picknik_ur_macro.xacro` and the end effector/payload into `picknik_ur_attachments_macro.xacro`. I also introduced a new macro per site config called `environment.xacro`, where the external geometry and cameras are defined. 

Finally, I introduced a convention for the macros to help with readability. Currently, the caller of the macro caller needs to know which links it exports to connect it to the robot or scene. I have added parent and child parameters so these values are explicitly set so there are no surprises.


src/picknik_ur_gazebo_config/config/config.yaml
```diff
@@ -61,14 +61,17 @@ hardware:
...
       - physical_parameters_file:
+          # Load physical_parameters.yaml from ur_description/config/<ur_type>
           package: "ur_description"
-          path: "config"
+          path: "config/%>> hardware.type <<%/physical_parameters.yaml"
       - visual_parameters_file:
+          # Load visual_parameters.yaml from ur_description/config/<ur_type>
           package: "ur_description"
-          path: "config"
+          path: "config/%>> hardware.type <<%/visual_parameters.yaml"
```
sea-bass, line 74: I dislike our reliance on this meta-language that parses YAML substitutions, but I suppose this is actually a bug fix in that prior to this, we were using a single set of defaults?

But if this change is not needed, would love to avoid it.

sea-bass (commented):
Big fan of this cleanup! Just one comment, really.

src/picknik_ur_gazebo_config/config/config.yaml
pac48: I am not a fan of either of the YAML substitutions. This logic can be moved to inside of the xacro, but I am also not a huge fan of of a lot of logic in xacro. I did it this way just because the base config already had it like this. I don't have a strong preference

sea-bass: I see, yeah let's keep it then.

pac48:
> Big fan of this cleanup! Just one comment, really.
> 
> EDIT: Also, seems like there are still some bugs to squash here, based on the CI output.

Just need to fix the gazebo camera now. I realized that it is published by the view is blocked.

sea-bass:
> Just need to fix the gazebo camera now. I realized that it is published by the view is blocked.

lol yes, this is really annoying and I found this out a while back. Funny enough, with the two Gazebo renderers (OGRE and OGRE2), this issue only happens with one renderer but not the other... but the "good" renderer also seems to have more trouble on non-x86 architectures.

pac48:
> > Just need to fix the gazebo camera now. I realized that it is published by the view is blocked.
> 
> lol yes, this is really annoying and I found this out a while back. Funny enough, with the two Gazebo renderers (OGRE and OGRE2), this issue only happens with one renderer but not the other... but the "good" renderer also seems to have more trouble on non-x86 architectures.

So is the correct solution enabling/disabling the `use_mesh` depending on the gazebo renderer? Currently, it just checks if it is Gazebo. If it is Gazebo, then it uses the cube to represent the camera, otherwise, it uses the realsense mesh.

pac48:
> > So is the correct solution enabling/disabling the `use_mesh` depending on the gazebo renderer? Currently, it just checks if it is Gazebo. If it is Gazebo, then it uses the cube to represent the camera, otherwise, it uses the realsense mesh.
> 
> Yep, that's how it was before this change, I believe.

I tried with both renders actually and neither works with the mesh enabled. I'm not sure how it was working before. Unless there is some combination of macro args that works. I thought all I would have to do is change the `gazebo_renderer` in the config.yaml. But that did not work.

src/picknik_ur_base_config/description/picknik_ur_macro.xacro
```diff
@@ -0,0 +1,89 @@
...
+                    <origin xyz="0.0 0.0 0.0" rpy="0.0 0.0 0.0"/>
+                </joint>
+            </xacro:if>
+        </xacro:unless>
+
+        <!-- pinch geometry for moveit motion planning. see https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/issues/611 -->
+        <xacro:if value="$(arg use_pinch_links)">
+            <xacro:forearm_pinch_link/>
+            <xacro:wrist_3_pinch_link/>
+        </xacro:if>
```
sea-bass, line 86: Should these be part of the attachments macro as well? At the least the `tool0` link definition should be, right?

src/picknik_ur_gazebo_config/description/environment.xacro
```diff
@@ -0,0 +1,134 @@
+<?xml version="1.0" encoding="utf-8"?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+  <!-- wrist_3_link/forearm_link have a radius 0f 0.0375m. Actual offset from surface is (radius - attached_link's radius). -->
+  <!-- UR Drivers check for a 2.5cm  distance between tool_flange/wrist_3 and forearm links. -->
```
sea-bass, line 4: same here re: these comments belonging with the UR/ UR attachments macro instead

src/picknik_ur_gazebo_config/description/environment.xacro
```diff
@@ -0,0 +1,134 @@
...
+          <remapping>/servo_controller/joint_trajectory:=/robot_controllers/joint_trajectory</remapping>
+        </ros>
+      </plugin>
+    </gazebo>
+
+    <gazebo>
+      <plugin filename="libignition-gazebo-sensors-system.so" name="ignition::gazebo::systems::Sensors">
+        <render_engine>${gazebo_renderer}</render_engine>
+      </plugin>
+    </gazebo>
```
sea-bass, line 131: same here re: these not being part of the environment, but of the overall top-level

src/picknik_ur_mock_hw_config/description/environment.xacro
```diff
@@ -0,0 +1,131 @@
+<?xml version="1.0" encoding="utf-8"?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+  <!-- wrist_3_link/forearm_link have a radius 0f 0.0375m. Actual offset from surface is (radius - attached_link's radius). -->
+  <!-- UR Drivers check for a 2.5cm  distance between tool_flange/wrist_3 and forearm links. -->
```
sea-bass, line 4: here too

src/picknik_ur_base_config/description/picknik_ur_attachments_macro.xacro
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+    <xacro:macro name="picknik_ur_attachments" params="parent child has_tool_changer">
+        <!-- parameters -->
+        <xacro:arg name="simulation" default=""/>
```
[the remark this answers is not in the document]
pac48: I think you might be referring to this a few lines below: `<xacro:property name="simulation" default="$(arg simulation)"/>` The simulation value is being set based on the args passed into xacro/config overrride

src/picknik_ur_base_config/description/environment.xacro
```diff
@@ -0,0 +1,110 @@
...
+    <joint name="external_camera_joint" type="fixed">
+      <parent link="${parent}" />
+      <child link="external_camera_link" />
+      <origin xyz="-0.3 0.3 1.0" rpy="0.0 0.4 0" />
+    </joint>
+
+    <xacro:realsense_d435 parent="external_camera_link" name="scene_camera" visible="false">
+      <origin xyz="0 0 0" rpy="0 0 0" />
+    </xacro:realsense_d435>
+  </xacro:macro>
```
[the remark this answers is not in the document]
pac48: I feel mixed on this one. For example, we might have an environment for our motion capture lab that would have many cameras. In that case, I would say the cameras are part of the environment. It would be easier to put a new robot in the same environment, without repeating the whole camera setup. For one camera, it seems easier to comprehend at the top level.

src/picknik_ur_base_config/description/picknik_ur_macro.xacro
pac48: Actually, tool0 is the link defined by the UR robot xacro, so we have no control over that. It is possible to put the tool changer in the attachments though if that makes sense.

sea-bass: Sorry, didn't realize `tool0` was the parent, not the child. I meant the child link definition. Pinch link you could argue could be a top level thing since it's an extra addition for motion planning to be well behaved.

src/picknik_ur_base_config/description/environment.xacro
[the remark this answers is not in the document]
sea-bass: Yeah, I conceptually agree, but was thinking more about separating the passive environment (collision objects, etc.) with the placement of external sensors.

It's also more visible to the user that you are explicitly putting a camera somewhere instead of it "just being part of the background".

That's why I was leaning top-level. But I defer to you!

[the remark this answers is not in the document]
pac48: Okay, I think it can be changed easy enough in the future if needed, so I will leave it for now

sjahr (commented):
Thanks for the cleanup!
When I try to launch the agent with these changes I get the following error:

```bash
Sourcing Overlay workspace.
Sourcing User workspace.
studio-user@hedgehog12:~/user_ws$ exit^C
studio-user@hedgehog12:~/user_ws$ robot.app ^C
studio-user@hedgehog12:~/user_ws$ agent_robot.app
[INFO] [launch]: All log files can be found below /home/studio-user/.ros/log/2024-02-29-12-08-27-433516-hedgehog12-165
[INFO] [launch]: Default logging verbosity is set to INFO
Replacing ${GAZEBO_RENDERER:-ogre} with ogre
Replacing ${UR_HARDWARE_TYPE:-ur5e} with ur5e
Replacing ${MOCK_HARDWARE:-true} with true
Loading behavior plugin from /home/studio-user/user_ws/src/moveit_studio_example_behaviors
name 'gazebo_renderer' is not defined
when evaluating expression 'gazebo_renderer'
Found Objective library directory at /home/studio-user/user_ws/install/picknik_ur_base_config/share/picknik_ur_base_config/objectives
Found Objective library directory at /home/studio-user/user_ws/install/picknik_ur_site_config/share/picknik_ur_site_config/objectives
Found Objective library directory at /home/studio-user/user_ws/install/picknik_ur_gazebo_config/share/picknik_ur_gazebo_config/objectives
Replacing ${GAZEBO_RENDERER:-ogre} with ogre
Replacing ${UR_HARDWARE_TYPE:-ur5e} with ur5e
Replacing ${MOCK_HARDWARE:-true} with true
Loading behavior plugin from /home/studio-user/user_ws/src/moveit_studio_example_behaviors
name 'gazebo_renderer' is not defined
when evaluating expression 'gazebo_renderer'
Found Objective library directory at /home/studio-user/user_ws/install/picknik_ur_base_config/share/picknik_ur_base_config/objectives
Found Objective library directory at /home/studio-user/user_ws/install/picknik_ur_site_config/share/picknik_ur_site_config/objectives
Found Objective library directory at /home/studio-user/user_ws/install/picknik_ur_gazebo_config/share/picknik_ur_gazebo_config/objectives
Replacing ${GAZEBO_RENDERER:-ogre} with ogre
Replacing ${UR_HARDWARE_TYPE:-ur5e} with ur5e
Replacing ${MOCK_HARDWARE:-true} with true
Loading behavior plugin from /home/studio-user/user_ws/src/moveit_studio_example_behaviors
name 'gazebo_renderer' is not defined
when evaluating expression 'gazebo_renderer'
Found Objective library directory at /home/studio-user/user_ws/install/picknik_ur_base_config/share/picknik_ur_base_config/objectives
Found Objective library directory at /home/studio-user/user_ws/install/picknik_ur_site_config/share/picknik_ur_site_config/objectives
Found Objective library directory at /home/studio-user/user_ws/install/picknik_ur_gazebo_config/share/picknik_ur_gazebo_config/objectives
Successful: Created and wrote system configuration: /home/studio-user/.config/moveit_pro/picknik_ur_gazebo_config/auto_created/complete_system_config.yaml
name 'gazebo_renderer' is not defined
when evaluating expression 'gazebo_renderer'

Failed to save robot description to file. Without this, the REST API won't work. Shutting down the agent.
```
I've already updated my ROS deps. Anything else I need to do to test this?

sjahr (approved):
Works for me now, thanks!

PR #227, opened 2024-02-27
Increase delays for p-stop test

sea-bass:
We have been seeing flakiness in the p-stop test, so cranking up some timeouts / delays.

NOTE: I had to fall to `--executor sequential` because the integration and p-stop tests were clashing with each other ☹️

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,7 +72,7 @@ jobs:
           . /opt/underlay_ws/install/setup.sh
           . /opt/overlay_ws/install/setup.sh
           . ./install/setup.sh
-          colcon test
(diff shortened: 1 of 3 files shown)


orensbruli:
> the integration and p-stop tests were clashing with each other
Q: Is this to be expected or is there something we would have to fix so that this doesn't happen? Or is it simply that colcon test is not intended for integration testing?

orensbruli (approved):

sea-bass:
I think the robot description topics in these integration tests are clashing with each other now that the `moveit_studio_ur_pstop_manager` tests are active for longer. I'm going to merge this and we probably want to keep investigating how to ensure the integration test packages don't run in parallel with each other without going full sequential.

PR #228, opened 2024-02-27
Move UR P-stop manager test to integration testing package

sea-bass:
This PR moves the `moveit_studio_ur_pstop_manager` integration test to the integration testing package so we can run `colcon test` without sequential execution.

Thoughts?


orensbruli:
LGTM, it's a cleaner solution to avoid collision between both tests.

PR #233, opened 2024-03-05
Fix timeout value

pac48:
The timeout parameter for the objectives should be a number.

### DIFF
--- a/src/picknik_ur_base_config/objectives/interpolate_to_joint_state.xml
+++ b/src/picknik_ur_base_config/objectives/interpolate_to_joint_state.xml
@@ -3,7 +3,7 @@
     <BehaviorTree ID="Interpolate to Joint State" _description="Move to a specified joint state using joint interpolation">
         <Control ID="Fallback" name="root">
             <Control ID="Sequence">
-                <Action ID="RetrieveJointStateParameter" timeout_sec="" joint_state="{target_joint_state}"/>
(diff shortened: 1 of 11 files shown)


sea-bass (approved):

sea-bass:
Gonna wait til the other upstream PR is fixed and run the integration tests on this PR, just to be safe.

sea-bass:
There seem to be some more failures 😢 

https://github.com/PickNikRobotics/moveit_studio_ur_ws/actions/runs/8158717690/job/22303589510#step:9:275

pac48:
> There seem to be some more failures 😢
> 
> https://github.com/PickNikRobotics/moveit_studio_ur_ws/actions/runs/8158717690/job/22303589510#step:9:275
> 
> EDIT: Wait, this is related to the same thing you just fixed, isn't it? Hmm...

Should https://github.com/PickNikRobotics/moveit_studio/blob/40cfa26abb9f36f2041bf3f07148ba3ea526947c/src/moveit_studio_behavior/config/tree_nodes_model.xml#L1367 be changed to 
```xml
<input_port name="constraints" default="{constraints}">Motion planning constraints.</input_port>
```

sea-bass:
So, I think before this latest change, there were some ports that were allowed to be empty -- this includes the timeout you changed in this PR, plus the constraints in the motion planning tasks that are still failing.

We should try find a way to continue allowing this in some way, since we should be able to do motion planning tasks without any actual constraints defined.

sea-bass (approved):

PR #235, opened 2024-03-06
Add request teleoperation for mock hw config

Abishalini:
To resolve the following error
image


sea-bass (changes requested):
Might be better to have this one live in the `picknik_ur_base_config` package, and then move the existing one to `picknik_ur_site_config`... this would help in the event we have more mock hw configs later.

sea-bass (approved):
Thanks!

PR #237, opened 2024-03-11
Remove the use of MoveToX Behaviors

Abishalini:
Resolves https://github.com/PickNikRobotics/moveit_studio/issues/6647

### DIFF
--- a/src/picknik_ur_base_config/objectives/request_teleoperation.xml
+++ b/src/picknik_ur_base_config/objectives/request_teleoperation.xml
@@ -26,7 +26,7 @@
             <Decorator ID="ForceSuccess" _while="teleop_mode == 4">
               <Control ID="Sequence">
                 <Action ID="SwitchUIPrimaryView" primary_view_name="Visualization"/>
-                <SubTree ID="Move to Pose"/>
+                <SubTree ID="Move to Pose" controller_names="/joint_trajectory_controller"/>
               </Control>
             </Decorator>
             <!-- Waypoint buttons move to joint state -->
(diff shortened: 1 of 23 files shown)


src/picknik_ur_base_config/objectives/request_teleoperation.xml
```diff
@@ -26,7 +26,7 @@
             <Decorator ID="ForceSuccess" _while="teleop_mode == 4">
               <Control ID="Sequence">
                 <Action ID="SwitchUIPrimaryView" primary_view_name="Visualization"/>
-                <SubTree ID="Move to Pose"/>
+                <SubTree ID="Move to Pose" controller_names="/joint_trajectory_controller"/>
```
sjahr, line 29: Is the hardcoded controller name necessary?

src/picknik_ur_site_config/objectives/request_teleoperation.xml
```diff
@@ -26,7 +26,7 @@
             <Decorator ID="ForceSuccess" _while="teleop_mode == 4">
               <Control ID="Sequence">
                 <Action ID="SwitchUIPrimaryView" primary_view_name="Visualization"/>
-                <SubTree ID="Move to Pose"/>
+                <SubTree ID="Move to Pose" controller_names="/joint_trajectory_controller"/>
```
sjahr, line 29: Same here, maybe add a comment why this is required

sjahr (commented):
Looks good just one thing :+1:

src/picknik_ur_base_config/objectives/request_teleoperation.xml
Abishalini: I made `controller_names` as a "param" for `Move to Pose` subtree because this subtree requires different controllers in other cases

sjahr: That makes sense, thanks!

sjahr (approved):

PR #240, opened 2024-03-18
Update planning scene topic and octomap collision check servo parameters

sea-bass:
This PR corrects the monitored planning scene being used for MoveIt Servo so that attached object collisions work.

It also adds the `check_octomap_collisions` parameter mostly for users to know this is also a new option.


src/picknik_ur_mock_hw_config/objectives/test_attach.xml
sea-bass, line 1: need to remove this file

sjahr (approved):

PR #243, opened 2024-03-20
Bump to version `5.0.0`


orensbruli:
This PR contains the needed changes to bump the version to the next minor release to main :eye: Please review that the changes are correct as the replacement of the version numbers is done automatically and can be wrong.


sea-bass (approved):
I'm sorry, I already cut a `v5.0` branch here, could you have it target that one instead (and then we'll merge that back to main)?

orensbruli:
@sea-bass done as you suggested.

sea-bass (approved):

PR #244, opened 2024-03-26
Change path to base Docker Compose with Deb installer changes

chancecardona:
This I think fixes some bugs that get masked a little bit if all your environment variables are set right. Also have a PR in the picknik ws.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -8,7 +8,7 @@ services:
     # Extend the installed MoveIt Pro docker compose file.
     # Change this to match your environment, if MoveIt Pro was installed to a different location.
     extends:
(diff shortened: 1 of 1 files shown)


orensbruli:
Is there any way I can it make intentionally fail to later check your fix?

chancecardona:
Simply test with moveit_pro PR https://github.com/PickNikRobotics/moveit_studio/pull/6928 which should expose this bug. My other PR's are also hammering away at this. Either way this is the correct variable now and this should go in, even if this does break the (soon unsupported) non-deb installs, it's a 1-line fix for them.

dyackzan (approved):

PR #246, opened 2024-03-27
Remove obsolete 'version' tag from docker-compose file

dyackzan:


chancecardona (approved):
Can we get one for the picknik_ur_ws now too please? Should be it for the main ones.

PR #252, opened 2024-04-04
Update camera launch logic

sjahr:
Address changes introduced by https://github.com/PickNikRobotics/moveit_studio/pull/6966

Requires also https://github.com/PickNikRobotics/picknik_ur5e_configs/pull/69

See https://github.com/PickNikRobotics/picknik_ut_moveit_studio_ws/pull/27 as a working reference

# TODO
- [ ] Test on HW
  - Decide when and where the cameras should be automatically launched


EzraBrooks (changes requested):
The serial number bug is specifically because of YAML coercing numbers that start with `0` into octal - so if we entirely eliminate the `cameras.yaml`, which is no longer needed, and just do this in XML launch files, it'll be very simple.

src/picknik_ur_base_config/launch/rs_cameras.launch.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<launch>
+  <arg name="scene_camera_serial_no" default="'0'" />
+  <arg name="wrist_mounted_camera_serial_no" default="'0'" />
+
+  <!-- Scene camera  -->
+  <include file="$(find-pkg-share realsense2_camera)/launch/rs_launch.py">
+    <arg name="camera_name" value="scene_camera"/>
+    <arg name="serial_no" value="$(arg scene_camera_serial_no)"/>
```
pac48, line 9:
```suggestion
    <arg name="serial_no" value="$(var scene_camera_serial_no)"/>
```

src/picknik_ur_base_config/launch/rs_cameras.launch.xml
```diff
@@ -0,0 +1,45 @@
...
+    <arg name="registered_rgb_depth_pair.depth_info" value="/scene_camera/color/camera_info"/>
+    <arg name="registered_rgb_depth_pair.depth_image" value="/scene_camera/depth/image_rect_raw"/>
+    <arg name="registered_rgb_depth_pair.registered_info" value="/scene_camera/color/camera_info"/>
+    <arg name="registered_rgb_depth_pair.registered_image" value="/scene_camera/depth/image_rect_raw"/>
+  </include>
+
+  <!-- Wrist camera  -->
+  <include file="$(find-pkg-share realsense2_camera)/launch/rs_launch.py">
+    <arg name="camera_name" value="wrist_mounted_camera"/>
+    <arg name="serial_no" value="$(arg wrist_mounted_camera_serial_no)"/>
```
pac48, line 28:
```suggestion
    <arg name="serial_no" value="$(var wrist_mounted_camera_serial_no)"/>
```

pac48:
I tried to test these PRs (https://github.com/PickNikRobotics/picknik_ur5e_configs/pull/69, https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/252) on the hardware and ran into a few issues. I added a suggested change for the camera launch file args. I think you need to use the `var` keyword instead of `arg`. 

Also, I had to install the  realsense2-camera package manually, so I think the package.xml needs to be updated. 

It worked after changing those things.

pac48 (dismissed):

pac48 (approved):

PR #253, opened 2024-04-05
Refactor objectives that still include the getPoseFromUser behavior

moth-man:
Refactors leftover objectives that still had behavior tree xmls that included the `getPoseFromUser` behavior and instead replaces them with the new sequence of `getPointsFromUser` -> `getPointCloud` -> `getPoseFromPixelCoords` -> `adjustPoseWithIMarker`

To test, first checkout [this branch](https://github.com/PickNikRobotics/moveit_studio/pull/6971) of `moveit_studio`, switch your user_ws in MoveIt Pro to point to the `moveit_studio_ur_ws` and test the following objectives:

- Teach Apriltag Grasp Offset (`picknik_ur_gazebo_config`)
- Teach Apriltag Grasp Offset (`picknik_ur_site_config`)
- Pick Object
- Pick and Place Object

Closes [#6871](https://github.com/PickNikRobotics/moveit_studio/issues/6871)


marioprats:
Tested this with Studio `main` and for some reason the `GetPointsFromUser` Behavior doesn't finish.

Teach Grasp Offset From AprilTag: After going through all steps (clicking object, confirming grasp), the Objective seems stuck in `GetPointsFromUser` and doesn't finish.
image

Pick Object: After clicking on the object and confirming the grasp, the Objective gets stuck in `GetPointsFromUser` 
image

moth-man:
@marioprats Sorry I forgot to mention that this needs to be tested with this [branch of studio](https://github.com/PickNikRobotics/moveit_studio/pull/6971) to test properly. I've updated the description for this PR to signify this

schornakj:
The situation with the TestPickObject integration test is a little complicated.
- The new version of the Pick Object objective generates the grasp pose by projecting the user's selected point into a point cloud, so it requires sensor data to be published while the objective is running.
- This is different from how it worked previously -- the earlier version of this objective just passed in the grasp pose directly.
- The integration tests all use the `picknik_ur_site_config` config package, which is a mock sim environment without simulated cameras.
- To run this objective, we need to be in an environment with simulated point cloud publishers. Either Gazebo or something to mock it.
  - There are some hurdles to overcome to get Gazebo running in CI, since Gazebo requires a display. One way around this is to use xvfb [like this](https://github.com/gazebosim/gz-rendering/blob/ign-rendering5/.github/ci/after_make.sh), but this requires more investigation.

I'm inclined to just disable the test in question for now and fix the integration test issue in a separate PR.

src/picknik_ur_site_config/objectives/pick_object.xml
```diff
@@ -3,17 +3,24 @@
...
         <Control ID="Sequence" name="root">
             <SubTree ID="Open Gripper"/>
-            <!-- Wrap in a ForceSuccess decorator so this can be used as both a top-level tree and subtree -->
-            <Decorator ID="ForceSuccess">
-                <Action ID="GetPoseFromUser" parameter_name="pick_object.grasp_pose" parameter_value="{grasp_pose}" />
-            </Decorator>
+            <Control ID="Sequence">
+                <Action ID="GetPointsFromUser" view_name="/wrist_mounted_camera/color/image_raw" point_names="grasp_pose" point_prompts="Select the object you wish to pick" pixel_coords="{pixel_coords}"/>
+                <Action ID="GetPointCloud" topic_name="/wrist_mounted_camera/depth/color/points" message_out="{point_cloud}"/>
+                <Action ID="TransformPointCloudFrame" output_cloud="{point_cloud}" target_frame="wrist_mounted_camera_color_optical_frame" input_cloud="{point_cloud}"/>
```
EzraBrooks, line 9: I think this needs to be added to the others as well

schornakj (approved):

PR #256, opened 2024-04-15
Use updated YAML-free version of SetupMTCUpdateGroupCollisionRule

schornakj:
Depends on https://github.com/PickNikRobotics/moveit_studio/pull/7086.

- Pass inputs to SetupMTCUpdateGroupCollisionRule as ports instead of though YAML parameter file.
- Delete `push_along_axis_config.yaml`, since its parameters are no longer needed.


pac48 (changes requested):
@schornakj I think both close_cabinet_door.xml  and push_button_ml.xml need to be updated as well.

pac48 (approved):

PR #271, opened 2024-05-06
Add reference application for bin picking

schornakj:
This adds a config package to demonstrate a bin picking application.

As configured in the `picknik_ur_bin_picking` package, the bins are placed next to each other near the robot. The system will pick objects from the origin bin and place them in the destination bin. Once all feasible objects have been picked, the bins assigned as origin and destination will be switched and the objective will repeat.


nbbrooks:
Closing this since we will do the approach that inherits from the base config

PR #275, opened 2024-05-16
Add multi-dof joint state to waypoints

Abishalini:


Abishalini:
The multi-dof field is optional and hence this change in not required

PR #279, opened 2024-05-23
Add VelocityForceController

dyackzan:


src/picknik_ur_base_config/config/control/picknik_ur.ros2_control.yaml
```diff
@@ -21,6 +21,8 @@ controller_manager:
       type: joint_trajectory_controller/JointTrajectoryController
     joint_trajectory_admittance_controller:
       type: joint_trajectory_admittance_controller/JointTrajectoryAdmittanceController
+    velocity_force_controller:
```
marioprats, line 24: Do we have to add the `velocity_force_controller` dependency to `package.xml`?

marioprats (approved):
Thanks David for all your work on this!

PR #295, opened 2024-06-26
Match casing of 'as' and 'FROM' in Dockerfile

dyackzan:
Gets rid of warning:
```
=> WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line xx)
```


chancecardona (approved):
Looks good

PR #300, opened 2024-06-28
Remove usage of YAML variable interpolators

EzraBrooks:
This PR clearly regresses some configurability of these packages. For example, non-mock-hardware-based configs will now need to explicitly set `simulated: false` in their `config.yaml`. A UR10 will have to explicitly set the UR hardware type. However, I think that verboseness is better in this context.

I'm not super concerned about hardcoding `ogre` as the rendering engine for Gazebo since `ogre2` seemingly works extremely inconsistently.


src/picknik_ur_base_config/config/config.yaml
```diff
@@ -11,13 +11,12 @@ hardware:
   # You can change this to another UR model but you must update any configuration affected by the different arm size.
   # UR models in the ur_description package are ur3, ur3e, ur5, ur5e, ur10, ur10e, and ur16e.
   # [Required]
-  type: ${UR_HARDWARE_TYPE:-ur5e}
+  type: ur5e
```
Abishalini, line 14: Do you think we can make this parameter not required? 
This parameter makes sense only for switching between different URs

MikeWrock: I agree we should drop this, not only is it UR specific but the UR macro uses the key `name` instead.

[In this example](https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/301/files) I've replaced the `picknik_ur` macro with the `ur_description`'s `ur_robot` macro so that I can use multiple different URs in my config. This is not possible using the `type:` tag in `config.yaml`

EzraBrooks: If we're going to make schema changes to the config.yaml like this (which I think we should do!) I'm going to go make a different PR in Pro that uses Pydantic to read and validate our YAML files so we have a better idea of the actual shape of the data before we start making changes.

Abishalini:
Created a PR targeting this branch which removes all the deprecated fields to get mock HW and gazebo config running - https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/315/

I haven't made changes to the other configs such as scan_and_plan

src/picknik_ur_gazebo_config/config/config.yaml
```diff
@@ -6,21 +6,16 @@
...
 
 # Optional parameters that can be read in your launch files for specific functionality
-optional_feature_params:
-  gazebo_world_name: "space_station_blocks_world.sdf"
-  gazebo_gui: False
-  gazebo_verbose: True
+# optional_feature_params:
+#   gazebo_world_name: "space_station_blocks_world.sdf"
+#   gazebo_gui: False
+#   gazebo_verbose: True
```
MikeWrock, line 12: Is this a placeholder of just a forgotten artifact?

MikeWrock (approved):

src/picknik_ur_gazebo_config/config/config.yaml
MikeWrock: @Abishalini if this needs to get removed please open up a new PR

PR #303, opened 2024-07-09
Fix typo in docker-compose.yaml

falfab:
This PR fixed a possible typo in the `docker-compose.yaml` file.

Currently the `USER_GID` argument is set from the `STUDIO_USER_UID` env variable. 
Probably this might be a typo and `STUDIO_USER_GID` might be the correct variable.


EzraBrooks (approved):
great catch, thanks!

PR #304, opened 2024-07-09
Add mobile ur config

sjahr:


src/picknik_ur_mobile_config/LICENSE
```diff
@@ -0,0 +1,25 @@
+Redistribution and use in source and binary forms, with or without
```
sjahr, line 1: Since we're including stl files from the [ridgeback repository](https://github.com/ridgeback/ridgeback_robot): Can we add them + do we need to adjust our license?

marioprats: The ridgeback repo doesn't have a license it looks like.
And we are just adding it as a submodule without modifying any source, so I don't think we need to do anything license-wise.

src/picknik_ur_mobile_config/README.md
```diff
@@ -0,0 +1,5 @@
+# picknik_ur_mobile_config
+
+A MoveIt Pro MuJoCo simulation for PickNik's Universal Robots (UR) arms.
```
marioprats, line 3:
```suggestion
A MoveIt Pro MuJoCo simulation of a UR5 arm on a mobile Ridgeback base.
```

marioprats (approved):

src/picknik_ur_mobile_config/description/picknik_ur.xacro
```diff
@@ -0,0 +1,80 @@
...
+    <limit effort="1000.0" lower="-5.0" upper="5.0" velocity="0.175" acceleration="10.0"/>
+    <dynamics damping="20.0" friction="500.0" />
+  </joint>
+
+  <link name="virual_rail_link_2"/>
+
+  <joint name="linear_y_joint" type="prismatic">
+    <parent link="virual_rail_link_2" />
+    <child link="ridgeback_base_link" />
+    <origin xyz="0 0 0" rpy="0 0 -1.5707963267948966" />
```
marioprats, line 60: I think this introduces a 90-def offset between the URDF and the Mujoco model.
What about specifying the translation axes for these two linear joints, e.g. `<axis xyz="0 1 0"/>`, and setting the origin to zero. That should make it match with Mujoco I think.

marioprats (changes requested):
Just tested this and I think we need to update the linear joints to match between Mujoco and the URDF.

marioprats (approved):

PR #305, opened 2024-07-10
Improve inline documentation for 3_waypoint_pick_and_place

davetcoleman:
Comments should not repeat the same wording as the Node name, e.g. ClearSnapshot should not be documented as "Clear the snapshot"


davetcoleman:
@MikeWrock do you know if our mock UR config is being replaced in the next release by Mujoco version? If so, I would like to copy these changes to that site config also.

MikeWrock (approved):

MikeWrock:
> @MikeWrock do you know if our mock UR config is being replaced in the next release by Mujoco version? If so, I would like to copy these changes to that site config also.

The mujoco config is not a replacement for mock hw, but now that you mention it the mock hw config only serves the purpose of showing the machine tending environment. Any mock hw configuration details can be just as easily demonstrated in the ur_base or ur_site config

davetcoleman:
My expectation is that for new users going through vanilla training, they will be in Mujoco not Mock, correct?

PR #306, opened 2024-07-10
Improve inline documentation for constrained_pick_place

davetcoleman:
Added comments explaining what everything does
Removed a subtree because it decreased readability and was confusing to me. Why was it named "TopLevelSequence"?

image

Also, do we really need RetryUntilSucessful? Do our planners not already retry, internally? I know vanilla MoveIt does. I think it makes us look bad. @marioprats


src/picknik_ur_mock_hw_config/objectives/constrained_pick_place.xml
```diff
@@ -1,21 +1,24 @@
...
-            <SubTree ID="Move to Waypoint" waypoint_name="Grasp Right" joint_group_name="manipulator" controller_names="/joint_trajectory_controller /robotiq_gripper_controller" planner_interface="moveit_default"/>
-          </Decorator>
-          <SubTree ID="Close Gripper"/>
-          <Decorator ID="RetryUntilSuccessful" num_attempts="3">
-            <SubTree ID="Move to Waypoint" controller_names="/joint_trajectory_controller /robotiq_gripper_controller" joint_group_name="manipulator" waypoint_name="Grasp Left" planner_interface="moveit_all_planners" constraints="{constraints}"/>
-          </Decorator>
-        </Control>
+        <!--Move to pick location-->
+        <SubTree ID="Open Gripper"/>
+        <!--Note: sometimes constraint planning can fail, so we add a retry decorator to improve success-->
```
davetcoleman, line 12: @marioprats i added this comment - does it make sense to you? do we really need this decorator?

MikeWrock, line 12:
```suggestion
        <!--Note: Sampling based planners can be non-deterministic. The retry decorator improves the likelihood of success-->
```

davetcoleman: Great fix!

davetcoleman:
@marioprats what im confused about is there already is a retry "attempts" functionality built into moveit at a deeper level, so it feels redundant to add this

[setNumPlanningAttempts](https://docs.ros.org/en/jade/api/moveit_ros_planning_interface/html/classmoveit_1_1planning__interface_1_1MoveGroup.html#a30bb16ad546acb7b2c95c42112e05a6a)

PR #307, opened 2024-07-10
Add useful description to cycle_between_waypoints.xml

davetcoleman:
The current description is a repeat of the Objective name, and so is useless.


davetcoleman:
I think this should also be copied to our Mujoco config

jeremydbronson (changes requested):
`Example of repeatidly moving ` should be `Example of repeatedly moving`

MikeWrock (dismissed):

davetcoleman:
> `Example of repeatidly moving ` should be `Example of repeatedly moving`

Thanks, there's no spell check in the Github code editor. I'll be more careful next time. Fixed!

jeremydbronson (approved):

PR #309, opened 2024-07-10
Add inline documentation to move_along_path

davetcoleman:
Explained what this all does.

Also, should the "Create example Cartesian reference trajecotry" be a subtree, rather than just a sequence?

### DIFF
--- a/src/picknik_ur_mock_hw_config/objectives/move_along_path.xml
+++ b/src/picknik_ur_mock_hw_config/objectives/move_along_path.xml
@@ -1,8 +1,9 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <root BTCPP_format="4" main_tree_to_execute="Move Along Path">
   <!--//////////-->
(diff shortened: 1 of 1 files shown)


davetcoleman:
I think this should also be copied to our Mujoco config

davetcoleman:
Seems like an unrelated bug in the build?

```
/__w/moveit_studio_ur_ws/moveit_studio_ur_ws/src/external_dependencies/gz_ros2_control/ign_ros2_control/src/ign_system.cpp:37:  "ignition/gazebo/components/JointVelocityReset.hh" already included at /__w/moveit_studio_ur_ws/moveit_studio_ur_ws/src/external_dependencies/gz_ros2_control/ign_ros2_control/src/ign_system.cpp:35  [build/include] [4]
    Category 'build/include' errors found: 1
    Total errors found: 1
```

MikeWrock (approved):

marioprats (approved):

PR #312, opened 2024-07-12
remap ports for move to waypoint

marioprats:
Closes https://github.com/PickNikRobotics/moveit_studio/issues/7973

### DIFF
--- a/src/picknik_ur_base_config/objectives/move_to_waypoint.xml
+++ b/src/picknik_ur_base_config/objectives/move_to_waypoint.xml
@@ -14,7 +14,7 @@
     <SubTree ID="Move to Waypoint">
       <input_port name="waypoint_name" default=""/>
       <input_port name="joint_group_name" default="manipulator"/>
-      <input_port name="use_all_planners" default="false"/>
+      <input_port name="planner_interface" default="moveit_default"/>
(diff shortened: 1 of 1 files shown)


src/picknik_ur_base_config/objectives/move_to_waypoint.xml
```diff
@@ -11,6 +11,11 @@
     </Control>
   </BehaviorTree>
   <TreeNodesModel>
-    <SubTree ID="Move to Waypoint"/>
+    <SubTree ID="Move to Waypoint">
+      <input_port name="waypoint_name" default=""/>
+      <input_port name="joint_group_name" default="manipulator"/>
+      <input_port name="use_all_planners" default="false"/>
```
sjahr, line 17: I think this option can be removed, since it is not used?

sjahr (commented):
Thanks for adding this. What to do with the `use_all_planners` option? Either you remove it or we expose the `planner_interface` port in this behavior

src/picknik_ur_base_config/objectives/move_to_waypoint.xml
marioprats: Good catch, I initially did this in `v5.3`, but it has changed now in `main`. Updating to set `planner_interface` instead.

sjahr (approved):

PR #317, opened 2024-07-29
Fix Gazebo Scan and Plan Config

Abishalini:
Resolves https://github.com/PickNikRobotics/moveit_studio/issues/7909


marioprats (approved):
Thanks Abi!

PR #319, opened 2024-07-30
add VFC example to mujoco config

marioprats:
- Adds the JTAC and VFC to the Mujoco controllers config.
- Removes `ft_frame` from the URDF. It doesn't look it's needed and bring in confusion (it's not aligned with the Mujoco sensor frame).
- Sets the `robotiq_ft_sensor` site to be the same as `attachment_site`, right at the tip of the arm where the sensor would be attached. This also makes it match `tool0` in the URDF.
- Makes Mujoco joints stronger for better tracking.
- Adds an example VFC Objective, with a corresponding 'starting' waypoint.

Closes https://github.com/PickNikRobotics/moveit_studio/issues/7952.


dyackzan (dismissed):
I'm able to run this objective without issue but it looks a little odd because the table in the visualation/robot URDF is in a different position than the one in the mujoco model. We can change that in a separate PR

marioprats:
> I'm able to run this objective without issue but it looks a little odd because the table in the visualation/robot URDF is in a different position than the one in the mujoco model. We can change that in a separate PR

Good point. I made the table position match in Mujoco and the URDF. But then the robot would collide at the initial configuration, so I moved the table 5cm down. I also added an extra velocity at the end of the example to move up away from the table, so that we don't leave the gripper in collision.

dyackzan (approved):
Awesome, that looks great

PR #328, opened 2024-08-06
Update mobile base config

mkhansenbot:
First PR to upstream changes from the training. The main change is adding the objectives, and changing the description of the config. There are a couple other changes that were in those files that I'm not sure if they should be included or not, please review and let me know. If not I'll remove them.

A 2nd PR will be needed to add the "wall" object that will be sprayed, as well as the remaining objectives (20-25) related to that.

### DIFF
--- a/src/picknik_ur_mobile_config/config/config.yaml
+++ b/src/picknik_ur_mobile_config/config/config.yaml
@@ -81,9 +81,3 @@ ros2_control:
   # Optionally configure remapping rules to let multiple controllers receive commands on the same topic.
   # [Optional, default=[]]
   controller_shared_topics: []
-
-# Octomap manager configuration parameters
-octomap_manager:
-  # Input point cloud topic name. The *output* point cloud topic published by
-  # the Octomap manager node is defined in sensors_3d.yaml.
-  input_point_cloud_topic: "/wrist_camera/points"
--- a/src/picknik_ur_mobile_config/package.xml
+++ b/src/picknik_ur_mobile_config/package.xml
@@ -3,7 +3,7 @@
   <name>picknik_ur_mobile_config</name>
   <version>5.1.0</version>
 
-  <description>MuJoCo simulation configuration package for Picknik's UR robot on a linear rail</description>
+  <description>MuJoCo simulation configuration package for Picknik's UR robot on a mobile base</description>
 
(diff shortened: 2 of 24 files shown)


src/picknik_ur_mobile_config/config/config.yaml
```diff
@@ -81,9 +81,3 @@ ros2_control:
   # Optionally configure remapping rules to let multiple controllers receive commands on the same topic.
   # [Optional, default=[]]
   controller_shared_topics: []
-
-# Octomap manager configuration parameters
```
mkhansenbot, line 85: Was this change supposed to be here or should I remove it?

src/picknik_ur_mobile_config/package.xml
```diff
@@ -20,8 +20,6 @@
   <exec_depend>picknik_accessories</exec_depend>
   <exec_depend>realsense2_camera</exec_depend>
   <exec_depend>realsense2_description</exec_depend>
-  <exec_depend>robotiq_description</exec_depend>
```
mkhansenbot, line 23: Were the robotiq files supposed to be removed? I'm assuming they're not needed. If they are I can change the commit to re-add them.

src/picknik_ur_mobile_config/config/config.yaml
marioprats: This was outdated, so yes, should be removed as you did.

src/picknik_ur_mobile_config/package.xml
marioprats: I think these are still used by the URDF to bring in meshes e.g. `package://robotiq_description/meshes/visual/2f_85/ur_to_robotiq_adapter.dae`, and by the controller config to enable opening/closing the gripper, so I think we should keep these.

marioprats (commented):
Thanks Matt! Two comments:
1) I think we are missing the `Move To a StampedPose` subtree. Solution 4 fails because of that.
2) Now I'm wondering if we should have `picknik_ur_mobile_config` be a 'base' config with some basic examples, and have another `picknik_ur_mobile_manipulation_training_config` that inherits from that, and contains all these Objectives. On the one hand it's another config to maintain, but on the other hand it would help having an isolated config we use for training, without polluting it with other Objectives. For instance, we may want to add perception and  pick & place Behaviors. Should everything be in the same config or should we have separated configs for the different courses?

PR #330, opened 2024-08-06
Add VFC helper subtrees and examples to mujoco config

marioprats:
Adds two subtrees:
* `Command VFC`.  Wrapper around `PublishVelocityForceCommand` to make it easy to send commands to the VFC and exit on some conditions (force exceeded, distance traveled).
* `Force Relaxation`. Helper subtree around the VFC to achieve a zero-force state, therefore 'relaxing' the force after contact is made. Useful as a last step in a stop-on-contact Behavior.

And two Objectives:
* `Move Until Contact Example`: Example of how to use `Command VFC` to implement a stop-on-contact behavior.
* `Move With Velocity And Distance Example`: Example of how to use `Command VFC` to achieve a velocity / force reference (slide on the table), and stop after a Cartesian distance has been travelled.

Makes use of message converter Behaviors to extract x,y,z from a `Pose` message.

Closes https://github.com/PickNikRobotics/moveit_studio/issues/8123
Closes https://github.com/PickNikRobotics/moveit_studio/issues/8133


dyackzan (approved):
Looks great! I was able to run it no problem

PR #331, opened 2024-08-06
Add training behaviors

marioprats:
Adds some packages used in training courses:
* `translate_pose_solution`: Solution to the training exercise to create a Behavior to translate a pose given x,y,z. Not build by default (COLCON_IGNORE).
* `service_interface`: A package with a `.srv` file defining a basic ROS 2 Service.
* `service_example`: A package defining the actual ROS 2 Service, using `service_interface`. Python code.
* `call_my_service_solution`: Solution to the training exercise to create a Behavior to call a ROS 2 Service. COLCON_IGNORE'd.


MikeWrock:
How do you feel about putting all of these in an "Example behaviors" folder? It seems a little out of place with the rest of the configs

marioprats:
> How do you feel about putting all of these in an "Example behaviors" folder? It seems a little out of place with the rest of the configs

Makes sense. I did that and updated the `package.xml` files a bit too.
I also removed the `COLCON_IGNORE` files to make sure those Behaviors are built in CI and therefore kept up to date.

marioprats:
The CI failure should be fixed after https://github.com/PickNikRobotics/moveit_pro_ci/pull/5

MikeWrock:
All of the behaviors run as expected, I will add my registration behaviors along with some objectives that use these and some usage examples in a separate PR

MikeWrock (approved):

PR #334, opened 2024-08-06
Merge main changes into v6.0

dyackzan:


src/picknik_ur_multi_arm_config/config/moveit/multi_arm_ur.srdf
```diff
@@ -182,27 +182,6 @@
...
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_robotiq_85_right_finger_link" reason="Never"/>
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_robotiq_85_right_finger_tip_link" reason="Never"/>
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_robotiq_85_right_inner_knuckle_link" reason="Never"/>
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_robotiq_85_right_knuckle_link" reason="Never"/>
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_tool_changer_link" reason="Adjacent"/>
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_ur_to_robotiq_link" reason="Adjacent"/>
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_wrist_1_link" reason="Adjacent"/>
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_wrist_2_link" reason="Adjacent"/>
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_wrist_3_link" reason="Adjacent"/>
-    <disable_collisions link1="first_wrist_3_pinch_link" link2="first_wrist_mounted_camera_link" reason="Never"/>
```
chancecardona, line 205: Are we sure we want to remove the pinch link SRDF collision exceptions? If anything we should probably enable the pinch link on the multi arm xacro and then reenable these. Thoughts @sjahr?

I'm not necessarily blocking this PR on that though.

src/picknik_ur_mujoco_config/config/config.yaml
```diff
@@ -42,6 +42,7 @@ objectives:
       - "moveit_studio::behaviors::MTCCoreBehaviorsLoader"
       - "moveit_studio::behaviors::ServoBehaviorsLoader"
       - "moveit_studio::behaviors::VisionBehaviorsLoader"
+      - "moveit_studio::behaviors::ConverterBehaviorsLoader"
```
marioprats, line 45: I think we'll need to be more selective on what we merge into branch 6.0.
The new `ConverterBehaviorsLoader` plugin is not in the MoveIt Pro 6.0 release, so this config won't work

marioprats (commented):
@dyackzan what exactly do we need on branch 6.0? See my comment. I think some of the changes would require updating MP as well.

src/picknik_ur_mujoco_config/config/config.yaml
marioprats: Actually I think we can just drop [da43687](https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/334/commits/da436875a9a78ddcd7fe9ec8482b6d777f3de877), so that we don't require the new message builder Behaviors. This branch without that specific commit seems to work.

PR #342, opened 2024-08-12
Mujoco release

MikeWrock:
I suggest cloning the latest main of [ros2_mujoco](https://github.com/PickNikRobotics/ros2_mujoco) and `moveit_pro build` with a 6.0.0-rc4 image until rc5 is cut with [this PR](https://github.com/PickNikRobotics/moveit_studio/pull/8228) merged


MikeWrock:
@marioprats do you have any guess why the `Move With Velocity and Distance Example` objective drifts to the side? I had to remove the distance criteria because I don't think we have a behavior to unpack a pose message for the script to evaluate. It is removed in this version

src/picknik_ur_base_config/objectives/close_gripper.xml
```diff
@@ -1,8 +1,8 @@
...
+<?xml version='1.0' encoding='UTF-8'?>
 <root BTCPP_format="4" main_tree_to_execute="Close Gripper">
-    <BehaviorTree ID="Close Gripper" _description="Close the gripper">
-        <Control ID="Sequence" name="root">
-            <Action ID="MoveGripperAction" gripper_command_action_name="/robotiq_gripper_controller/gripper_cmd" position="0.7929"/>
-        </Control>
-    </BehaviorTree>
+  <BehaviorTree ID="Close Gripper" _description="Close the gripper">
+    <Control ID="Sequence" name="root">
+      <Action ID="MoveGripperAction" gripper_command_action_name="/robotiq_gripper_controller/gripper_cmd" position="0.7929" timeout="15.000000"/>
```
MikeWrock, line 5: This would sometimes take a bit longer than 10s and fail the integration test

src/picknik_ur_mujoco_config/description/ur5e.xml
```diff
@@ -284,7 +286,8 @@
   </worldbody>
 
   <actuator>
-    <position class="robot" name="linear_rail_joint" joint="linear_rail_joint" kp="75000" dampratio="1.0"/>
+    <!-- <general class="robot" name="linear_rail_joint" joint="linear_rail_joint" gainprm="25000" biasprm="0 -25000 -20000"/> -->
```
MikeWrock, line 289:
```suggestion
```

marioprats:
> @marioprats do you have any guess why the `Move With Velocity and Distance Example` objective drifts to the side? I had to remove the distance criteria because I don't think we have a behavior to unpack a pose message for the script to evaluate. It is removed in this version

The distance criteria is the stop condition. It will keep moving forever without that.
We should have the unpack Behaviors in MP `main`, not in the `6.0` release.
If this is for the `6.0` release I suggest just removing those example VFC Objectives.

MikeWrock:
> > @marioprats do you have any guess why the `Move With Velocity and Distance Example` objective drifts to the side? I had to remove the distance criteria because I don't think we have a behavior to unpack a pose message for the script to evaluate. It is removed in this version
> 
> The distance criteria is the stop condition. It will keep moving forever without that.
> We should have the unpack Behaviors in MP `main`, not in the `6.0` release.
> If this is for the `6.0` release I suggest just removing those example VFC Objectives.

@marioprats The fact that I didn't stop was not surprising, but it seems like it was moving in an odd coordinate frame, is it supposed to move in world? It looked like it was moving down into the left in the base frame but maybe it was moving in the wrist frame

MikeWrock:
> > > > @marioprats do you have any guess why the `Move With Velocity and Distance Example` objective drifts to the side? I had to remove the distance criteria because I don't think we have a behavior to unpack a pose message for the script to evaluate. It is removed in this version
> > > 
> > > 
> > > The distance criteria is the stop condition. It will keep moving forever without that.
> > > We should have the unpack Behaviors in MP `main`, not in the `6.0` release.
> > > If this is for the `6.0` release I suggest just removing those example VFC Objectives.
> > 
> > 
> > @marioprats The fact that I didn't stop was not surprising, but it seems like it was moving in an odd coordinate frame, is it supposed to move in world? It looked like it was moving down into the left in the base frame but maybe it was moving in the wrist frame
> 
> It should move down a couple of cm until it finds the table, then left, like this: move_with_velocity_and_stop 
This is what I was seeing:
vfc
But then I realized what the intention was and updated the start waypoint:
vfc2

All good now!

src/picknik_ur_mujoco_config/config/config.yaml
```diff
@@ -42,7 +42,6 @@ objectives:
       - "moveit_studio::behaviors::MTCCoreBehaviorsLoader"
       - "moveit_studio::behaviors::ServoBehaviorsLoader"
       - "moveit_studio::behaviors::VisionBehaviorsLoader"
-      - "moveit_studio::behaviors::ConverterBehaviorsLoader"
```
marioprats, line 45: In `main` we should be able to keep `ConverterBehaviorsLoader`, which is needed for the latest VFC examples

src/picknik_ur_mujoco_config/objectives/command_vfc.xml
```diff
@@ -13,17 +13,9 @@
             <Action ID="CreateStampedPose" reference_frame="{tip_link}" stamped_pose="{current_pose}"/>
             <Action ID="TransformPoseFrame" output_pose="{current_pose}" input_pose="{current_pose}"/>
             <Action ID="CalculatePoseOffset" source_pose="{initial_pose}" destination_pose="{current_pose}" source_to_destination_pose="{initial_pose_current}"/>
-            <Action ID="UnpackPoseMessage" pose="{initial_pose_current}"/>
```
marioprats, line 16: For `main` I think we should revert the changes to this file and the other VFC example Objectives, since the subtree won't work as expected without this.
When we merge into `6.0` we can remove this entire subtree and the two latest VFC examples: `Move Until Contact Example` and `Move With Velocity and Stop Distance`. I think it's okay to have those in the `6.1` release, and they are not needed for training.

marioprats (commented):
In `main` we should have access to the message builder Behaviors and should be able to keep the stop condition on the VFC example and subtree.
I think we should remove those when merging into the `6.0` branch.

marioprats:
@MikeWrock I added back the stop conditions and tested in `main`, everything seems to be working.
I would suggest merging this in `main`, and then cherry-pick it in `6.0` and remove the VFC Objectives and `ConverterBehaviorsLoader`.

marioprats (approved):

EzraBrooks:
I think this work got done elsewhere?

PR #352, opened 2024-08-22
Add airfoil to config

mkhansenbot:
Adds the airfoil to the mobile_manipulation_config


marioprats:
@mkhansenbot you needed to update the package your are loading the xacro from in the `config.yaml`, to point to this package. The airfoil rendered after doing that.
I added some more changes to make this work after making the base shorter, and to apply some other changes we had applied to the base config (stiffer joints, fixing joint ordering, using keyframe).
A better solution would be to have this package reuse the base package xacro by including it and just adding the airfoil. We can do that in a follow-up PR.

Also, you'll have to skip Objective 17, or remove it from the config, since it doesn't work as expected, but fixing it requires some changes in the backend.

marioprats:
I also rebased this on `v6.0` so that you can use it safely with the 6.0 release.

MikeWrock:
@mkhansenbot I'm out of office today but I will merge this tomorrow

MikeWrock:
I have a couple suggestions I will open up in a separate PR so this is not blocked. Some notes that I have:

- This looks really well put together, the objectives are straightforward and well laid out, and it really shows off the functionality of Pro
- Despite Mario's warning I ran objective 17 and it got stuck in a collision state and I had to reset. Objective 18 (and some other objectives that use ExecuteFollowJointTrajectory) is missing an ActivateControllers if it wasn't ran at some point in a previous objective. Should we add it to all the objectives that require it?
- I added a TransformPose to 03 - Pose 2m above the robot, I think it fits well and is not until #23
- I added an extra pose to 11 - Draw a square, so the visualize path draws a square (it does not draw the "from current" part of the path)
- I think 22 - Visualize airfoil frame should have a little more visualization, either load both point clouds (the guess and the registered one) in two different colours, visualize the pose of the before and after frames, or use visualize path (between the guess and aligned pose) and a pose marker to show the correction. Which one do you prefer?

Last thing is I noticed the picknik_ur_mobile_config is not removed, should we delete that?

MikeWrock (approved):

MikeWrock (approved):

PR #313, opened 2024-07-12
Copy example Objectives from Mimic Sim to Mujoco, with latest comments

davetcoleman:
Copy in these PRs to new Mujoco config
https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/308
https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/307
https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/306
https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/305

PR #357, opened 2024-08-29
Copy example Objectives from Mimic Sim to Arm On Rail Sim

bgill92:
Resolves #313.

This PR ports the basic Objectives from `picknik_ur_mock_hw_config` to `arm_on_rail_sim`. The waypoints were also cleaned up. The `Constrained Pick and Place` Objective was updated to account for now actually repeatedly picking and placing a physical object.

### To test:
1) Bring up MoveIt Pro with this config.
2) Run the following Objectives and ensure that they succeed for atleast 1 cycle:
```
3 Waypoints Pick and Place
Constrained Pick and Place
Cycle Between Waypoints
Joint Diagnostic
```


davetcoleman (approved):
I compared against my old PR https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/313/files

and everything looks great - comments and descriptons were preserved. thanks for working on this!!

davetcoleman:
Note i did not / cannot test on my machine, but i think this is ready for merge

PR #358, opened 2024-09-02
Add nvidia container toolkit to Dockerfile

nbbrooks:
Mujoco camera performance with an Nvidia GPU inside a Docker container is really bad. I am hoping this will fix it. Taken from Chance's BMW workspace PR https://github.com/PickNikRobotics/bmw_ws/pull/1.


nbbrooks:
Since I have not had much luck getting this configured, I tried a simpler workflow (no CUDA, just nvidia drivers) and got the render speed up I wanted via https://github.com/PickNikRobotics/moveit_studio_ur_ws/tree/pr-add-dockerfile-with-nvidia-drivers

Unfortunately, I do not think we can just merge the nvidia-driver change into the main Dockerfile. Ironically, not because it will break things for non-nvidia users (I used the image built with nvidia drivers on an Intel integrated graphics laptop successfully), but because it will cause a segfault when launching Pro if the naive version installed by the Dockerfile (e.g. 555) does not match the version on your system (e.g. 535).

Let's talk about how we want to handle this - I don't love having to keep two Dockerfiles up to date. Perhaps initially, unless there is something simple I missed with nvidia-toolkit, we could just add the commented out entries for nvidia-driver installation to the primary Dockerfile with instructions in the Quick Start to uncomment the lines and sync the driver version if you have an Nvidia card and want high render rates in the Pro sim.

Dockerfile
```diff
@@ -49,6 +49,25 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
       /home/${USERNAME}/.ros && \
     chown -R $USER_UID:$USER_GID /home/${USERNAME} /opt/overlay_ws/
```
chancecardona, line 51: For this image we also need to install the nvidia-container-toolkit:
```suggestion
RUN curl -fsSL https://nvidia.github.io/libnvidia-container/gpgkey | sudo gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg \
  && curl -s -L https://nvidia.github.io/libnvidia-container/stable/deb/nvidia-container-toolkit.list | \
    sed 's#deb https://#deb [signed-by=/usr/share/keyrings/nvidia-container-toolkit-keyring.gpg] https://#g' | \
    sudo tee /etc/apt/sources.list.d/nvidia-container-toolkit.list


```

Dockerfile
```diff
@@ -49,6 +49,25 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
...
 
+ENV DEBIAN_FRONTEND=noninteractive
+# Add wget
+# hadolint ignore=DL3008
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt,sharing=locked \
+    apt-get update && apt-get install --no-install-recommends -y software-properties-common wget
```
chancecardona, line 57:
```suggestion
    apt-get update && apt-get install --no-install-recommends -y software-properties-common wget nvidia-container-toolkit
```

chancecardona (changes requested):

EzraBrooks:
Unsure about how general of a fix it is, but I was able to help another customer get much better performance just by changing `runtime:` to `nvidia` in the docker-compose file, without modifying the Dockerfile.

chancecardona:
That is very interesting.

Yeah I think the solution to this is multistepped.

We probably need to provide a /etc/docker/daemon.json configuring the nvidia and cri-o options needed, in addition to the runtime in the docker compose as Ezra just mentioned. (nvidia-ctk can do some of this)

Overall I think the solution to this is to have: 
- nvidia container toolkit installed on the host
- nvidia drivers installed on host (cuda toolkit not required)
- configuring the files / daemons right (mentioned above)

chancecardona:
Here is an example of my ZFS + cuda enabled docker daemon.json:
```
{
    "default-runtime": "nvidia",
    "exec-opts": [
        "native.cgroupdriver=cgroupfs"
    ],
    "features": {
        "cdi": true
    },
    "runtimes": {
        "nvidia": {
            "args": [],
            "path": "nvidia-container-runtime"
        }
    },
    "builder": {"Entitlements": {"security-insecure": true }},
    "storage-driver": "zfs"
}
```

PR #359, opened 2024-09-03
add subtree_only to a couple of objectives

marioprats:
Closes https://github.com/PickNikRobotics/moveit_studio/issues/8507 and https://github.com/PickNikRobotics/moveit_studio/issues/8506.


EzraBrooks (changes requested):
Shouldn't this be going into v6.1?

EzraBrooks (approved):
oops, there isn't a v6.1 yet

PR #360, opened 2024-09-03
remove Cycle Between Waypoints from base config

marioprats:
Closes https://github.com/PickNikRobotics/moveit_studio/issues/8505


jeremydbronson (approved):

EzraBrooks (approved):
well there is no v6.1 yet..

PR #361, opened 2024-09-03
Add commented out Dockerfile commands for installing Nvidia drivers

nbbrooks:
Add instructions to Dockerfile for manual Nvidia driver installation until we figure out the X11 forwarding issue.

We will also update the Quick Start to make Nvidia GPU users aware of this.


Dockerfile
```diff
@@ -49,6 +49,15 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
       /home/${USERNAME}/.ros && \
     chown -R $USER_UID:$USER_GID /home/${USERNAME} /opt/overlay_ws/
 
+# IMPORTANT: Optionally install Nvidia drivers for improved simulator performance with Nvidia GPUs.
+# To do this you must
+# 1. Uncomment the ENV and RUN entries below
+# 2. Replace the 'nvidia-driver-555' apt package with the Nvidia driver version on your host, e.g. nvidia-driver-535, nvidia-driver-555. Use nvidia-smi on your host to determine the driver version.
+# After rebuilding via `moveit_pro build` verify the drivers are active in your container by running `nvidia_smi` inside of `moveit_pro shell`.
+# ENV DEBIAN_FRONTEND=noninteractive
+# RUN apt update && apt install -y software-properties-common
```
EzraBrooks, line 58: Aren't these two things already done?

Dockerfile
```diff
@@ -49,6 +49,15 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
...
+# RUN add-apt-repository ppa:graphics-drivers/ppa && apt update && apt upgrade -y && apt install -y nvidia-driver-555
```
EzraBrooks, line 59: Make sure to use the Apt cache mounts on any apt installation command

EzraBrooks (changes requested):

Dockerfile
MikeWrock, line 59:
```suggestion
# RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
#     --mount=type=cache,target=/var/lib/apt,sharing=locked \ 
#    add-apt-repository ppa:graphics-drivers/ppa && \
#    apt update && apt upgrade -y && apt install -y nvidia-driver-555
```

Dockerfile
nbbrooks: Hmmm...I do see them in the moveit_studio/Dockerfile (isn't that where `FROM ${MOVEIT_STUDIO_BASE_IMAGE} AS base` comes from)?

BUT, yesterday I removed the `ENV DEBIAN_FRONTEND=noninteractive` to see what would happen and it seemed to be waiting for user input while installing the drivers about keyboard language.

Dockerfile
nbbrooks: thanks Mike!

Dockerfile
nbbrooks: > ENV DEBIAN_FRONTEND=noninteractive
Chat GPT is telling me ARGs are not inherited from the base image, which would be consistent with my experience.

> RUN apt update && apt install -y software-properties-common
I think software-properties-common should be installed by the base image. But I think verifying it here should be really fast so I'd prefer to keep it in case someone needs to copy paste this hack into a diff Dockerfile where that might not be true.

MikeWrock (approved):

Dockerfile
EzraBrooks: > Chat GPT is telling me ARGs are not inherited from the base image, which would be consistent with my experience.

yeah but it's an ENV, not an ARG. ChatGPT is lying to you as usual ;)

but if you want to leave it here for copy pasta, ok

PR #363, opened 2024-09-09
Fix for gazebo configs

MikeWrock:
I'm not sure there is any way to prevent the linter from breaking a line like this in some other config, so I think maybe we just dont lint xacros?

### DIFF
--- a/src/picknik_ur_gazebo_config/description/picknik_ur_gazebo.xacro
+++ b/src/picknik_ur_gazebo_config/description/picknik_ur_gazebo.xacro
@@ -125,8 +125,7 @@
       name="ign_ros2_control::IgnitionROS2ControlPlugin"
     >
       <parameters>
-        $(find
-        picknik_ur_gazebo_config)/config/control/picknik_ur5e.ros2_control.yaml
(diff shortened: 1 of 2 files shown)


src/picknik_ur_gazebo_config/description/picknik_ur_gazebo.xacro
```diff
@@ -125,8 +125,7 @@
       name="ign_ros2_control::IgnitionROS2ControlPlugin"
     >
       <parameters>
-        $(find
-        picknik_ur_gazebo_config)/config/control/picknik_ur5e.ros2_control.yaml
+        $(find picknik_ur_gazebo_config)/config/control/picknik_ur5e.ros2_control.yaml
```
dsobek, line 128: Maybe we can get the prettier to skip over this line using ignore comments?
```
<!-- prettier-ignore-start -->
$(find picknik_ur_gazebo_config)/config/control/picknik_ur5e.ros2_control.yaml
<!-- prettier-ignore-end -->
```

https://github.com/prettier/plugin-xml?tab=readme-ov-file#ignore-ranges

dsobek (approved):

PR #368, opened 2024-09-11
Add airfoil to mobile_manipulation config

mkhansenbot:
This was previously merged to v6.0 but not to main, need to merge to main so that it will be included in future releases


src/mobile_manipulation_config/package.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" ?>
 <package format="3">
   <name>mobile_manipulation_config</name>
-  <version>5.1.0</version>
+  <version>6.0.0</version>
```
MikeWrock, line 4:
```suggestion
  <version>6.1.1</version>
```

MikeWrock (approved):

PR #370, opened 2024-09-18
Reduce build time and chance of OOM by removing unnecessary sources dependencies

nbbrooks:
Jeremy ran into a OOM error when testing 6.2-rc1 with the ur_ws on a VM with 8GB RAM. This prompted us doing an audit of our source dependencies.

 
Removes gz_ros2_control
- Our fork seems to be about adding mimic joint support
- It looks like it was added a diff way via https://github.com/ros-controls/gz_ros2_control/pull/33
- If ^ doesn't work, we already put a disclaimer in the tutorials referencing the Gz configs that they may not work. Ideally we will finish arm_on_rail_sim lab world migration for 6.3 and move the Gz configs out of the main ur_ws repo.  

Removes LMS1xx
- Small library, but throws annoying compilation warnings
- Don't seem to be using it anywhere
- There is a ros humble apt package for it

See https://picknik.slack.com/archives/C01HWFR3KAB/p1726688884384939 for more context

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,14 +10,6 @@
 	path = src/external_dependencies/serial
 	url = https://github.com/tylerjw/serial.git
 	branch = ros2
-[submodule "src/external_dependencies/gz_ros2_control"]
-	path = src/external_dependencies/gz_ros2_control
-	url = https://github.com/PickNikRobotics/gz_ros2_control.git
-	branch = mimic-joint-feature-humble
-[submodule "src/external_dependencies/LMS1xx"]
-	path = src/external_dependencies/LMS1xx
-	url = https://github.com/clearpathrobotics/LMS1xx.git
-	branch = humble-devel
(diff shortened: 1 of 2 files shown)


jeremydbronson (approved):

nbbrooks:
Going to hold off merging this until Jeremy has a chance to test if the Gz configs referenced in our tutorials
- launch successfully
- open/close gripper objectives still work (the gripper opens/closes)
- but not testing/worrying about grasping physics

mostly so we know the state of things for customer support requests.

nbbrooks:
@jeremydbronson ping on the Gz sanity check test.

jeremydbronson:
@nbbrooks I ran through this yesterday without issue.

nbbrooks:
@EzraBrooks please make sure this gets included in the 6.2 tag of ur_ws.

PR #372, opened 2024-09-20
Fix non unique cgroup validation error

chancecardona:
Fixes https://github.com/PickNikRobotics/moveit_studio/issues/8659

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -22,12 +22,7 @@ services:
   agent_bridge:
     extends: base
     privileged: true
-    # Allow the user to use the host's network video4linux and usb_device devices.
-    # "c 81:* rmw": This rule specifies that the container has read, write, and mknod access (rmw) to all character devices (c) with a major number of 81 (video4linux).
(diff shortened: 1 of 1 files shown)


nbbrooks:
Tagging @EzraBrooks for his history on this. Seems like cgroup is unnecessary since we are running privileged. But I am still unclear on how someone would encounter this, and how close we were to running unprivileged (Which this would move us farther away from).

chancecardona:
@nbbrooks it's not that we're removing the permissions, it's that we're removing the re-definition since it's already here: https://github.com/PickNikRobotics/moveit_studio/blob/3fb51807978711f7ca7c41c98fd424b80180fcc7/docker-compose-prod.yaml#L82

EzraBrooks:
I believe the cgroup was necessary - I don't recall exactly why. I think it had to be added when we switched to non-root usership in the container. Giving the container `privileged: true` doesn't necessarily mean you're giving _every user in the container_ root privileges.

EzraBrooks (approved):

PR #377, opened 2024-09-27
Fix realsense launch

dyackzan:


pac48 (approved):

nbbrooks:
I was going to pull this down on 007 to test but saw your message on slack you tested in on 002.

My Q about this is - it looks like the only consequential change in this pr is adding the `base_objectives` - why did we have to do that?

This is now the only config that has the `base_objectives` entry, but it is NOT our only config that inherits from a diff config. Do those other configs need `base_objectives` as well?

dyackzan:
Good question, I had to do that to get teleop working on the hardware but I'm unsure what made it necessary. I'll investigate that more tomorrow.
The only other consequential change that took a bit of digging was the change to the camera launch file [here](https://github.com/PickNikRobotics/moveit_studio_ur_ws/pull/377/files#diff-454c5b071ead9bc8183ea803a5c6c0ba3782f7bf946c030bbdceba0715806915R62-R66). Realsense updated the interface and now the `launch_setup()` method has a required `params` argument

dyackzan (commented):
Yeah @nbbrooks it looks like that really shouldn't be necessary. But for some reason the "Request Teleoperation" objective doesn't get passed through to the 002 hardware config, but all the other ones from the picknik_ur_base_config do...
Screenshot from 2024-09-30 17-39-14

dyackzan:
I created another issue to track the teleoperate error: https://github.com/PickNikRobotics/moveit_studio/issues/8789

I think this one is good to go now.

nbbrooks (approved):

nbbrooks:
Agreed the removed controllers should have been in a diff PR (and would have been a good cleanup change). But I merged it anyway :|

PR #378, opened 2024-10-01
Update mobile manipulation config objectives (fixes #8763)

mkhansenbot:
This fixes https://github.com/PickNikRobotics/moveit_studio/issues/8763

The issue is that the `LoadPointCloudFromFile` can't resolve a path with `~/user_ws/` so a package relative path is used instead. This puts `mobile_manipulation_config` into the `{package_name}` blackboard variable, and passes that to the behavior.


nbbrooks:
I'm adding @marioprats to this as he is looking into if there was a regression in being able to specify relative filepaths without a package name, e.g. ~/user_ws/xyz, which is what seems to have broken the tutorial objective.

PR #381, opened 2024-10-09
Removing primary view switch in teleop Joints Jog

Hoffalypse:
This is a removal of two `switchUIPrimaryView' behaviors. They are both tied to the `RequestTeleoperation` subtree. They are being removed so that in multi pane view when moving the sliders the view does not jump back to the main pane.

### DIFF
--- a/src/picknik_ur_base_config/objectives/request_teleoperation.xml
+++ b/src/picknik_ur_base_config/objectives/request_teleoperation.xml
@@ -153,10 +153,10 @@
               />
             </Control>
             <Control ID="Sequence" _while="teleop_mode == 1">
-              <Action
+              <!-- <Action
                 ID="SwitchUIPrimaryView"
                 primary_view_name="Visualization"
-              />
+              /> -->
               <Action
                 ID="TeleoperateJointJog"
                 controller_name="servo_controller"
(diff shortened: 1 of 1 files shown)


MikeWrock:
One thing I noticed when testing this out is the joint jog panel does not render over the BT pane, so if it is selected as the primary view then the joint jog panel does not appear. The previous requirement to change primary panes to Visualization masked this issue, but now it is not longer prevented.

I would say that the issue I noted is less painful than the panes jumping around on you, but I would like to hear @Hoffalypse thoughts on how we can either:
- Force the visualization pane to be primary **only** when the BT pane is currently the primary pane
- Render joint jog over the BT pane
- Pop a warning that joint jog is not shown when BT is the primary pane (and tell the user to change the primary pane to something else)

mewtwoshaurd (approved):
Played around with it, the sliders are kind of in the way of the robot but they don't jump around and the scene seems stable. Resized my window a few times too and had no issues

Hoffalypse:
@MikeWrock This will be fixed in #9069. The pane will not jump but all functionality will be shown in the visualization pane no matter where it exists now. I think this creates the best UX as I generally feel the less jumping the better.

Hoffalypse:
@mewtwoshaurd are you using a recent version of main. They shouldnt jump when switched anymore.

mewtwoshaurd:
@Hoffalypse this is on the most recent main branch I just build:

https://github.com/user-attachments/assets/a4a1b06e-063d-4d7f-b802-e2c213e86b86

Hoffalypse:
OK I may have spoke too soon, I apologize, it was in the merge Queue. Should be GTG now.

mewtwoshaurd (approved):
Double checked with new main and seems to work well -- I think I encountered a different UI bug that I'll make a ticket for, but these changes work

PR #383, opened 2024-10-09
Fix planner_interface Ports and other Bugfixes from Main

chancecardona:
Not doing to main as https://github.com/PickNikRobotics/moveit_studio/pull/8898 requires changes not in 6.2.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -14,7 +14,7 @@ services:
       target: user-overlay
       args:
         - USER_UID=${STUDIO_USER_UID:-1000}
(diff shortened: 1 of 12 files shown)


docker-compose.yaml
```diff
@@ -14,7 +14,7 @@ services:
       target: user-overlay
       args:
         - USER_UID=${STUDIO_USER_UID:-1000}
-        - USER_GID=${STUDIO_USER_UID:-1000}
+        - USER_GID=${STUDIO_USER_GID:-1000}
```
MikeWrock, line 17: Good catch!

MikeWrock (approved):

PR #386, opened 2024-10-16
Remove colcon build step from Dockerfile

MikeWrock:
You shouldn't have to build your workspace to get a docker image, and this step is super frustrating if you have broken packages but no way to quickly build them.

With this line removed, you can build a Pro image with a workspace that has broken packages. Then using `moveit_pro build user_workspace` you can more easily iterate on and fix your broken package

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -119,10 +119,6 @@ ENV USER_WS=${USER_WS}
 
 # Compile the workspace
 WORKDIR $USER_WS
-# hadolint ignore=SC1091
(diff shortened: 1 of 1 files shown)


chancecardona (approved):
This seems good to me. I think we originally did this when we had hardcoded entrypoints, which is no longer the case.

mewtwoshaurd:
Closes PickNikRobotics/moveit_studio#8463, because the `colcon-defaults.yaml` in the user_ws `colcon build` works fine but the Dockerfile `colcon build` doesn't have that yaml file access!

PR #390, opened 2024-10-18
Update Submodules

chancecardona:
Updates the commits to latest main for picknik_accessories, picknik_registration, ros2_robotiq_gripper  since we just merged some more PRs into them and they were on old commits or branches anyways.


chancecardona:
We could repoint our repo to humble, but why are there 2 branches (main and humble) for the robotiq_gripper if it's our fork? Because of that I think we need to fix this package on main, even if it's not a true `rolling` main.

pac48 (approved):

PR #395, opened 2024-11-12
fix: subcategories for arm_on_rail and base picknik behaviors

mewtwoshaurd:
just adding subcategories to the ur_ws behaviors and renaming the `Wrist` and `Scene` camera objectives per @davetcoleman request


davetcoleman (dismissed):
Excellent work. Chatted with shaur on this in two video calls

davetcoleman (approved):
Discussed on call

PR #396, opened 2024-11-14
Fix register cad part objective

dsobek:
It looks like some blackboard names changed. Also some subtrees now have a _collapsed input port that I removed for subtrees of this objective.


shaur-k (changes requested):
Couldn't get `register_cad_part` to work -- looks like you may have to change the name of an output port too somewhere? 
image

mewtwoshaurd:
it's a very quick fix -- the "CreateStampedPose" Initial Guess block needs to output `pose_stamped` as well -- I'll make those changes and commit it here

shaur-k (approved):

MikeWrock (approved):

PR #402, opened 2024-11-21
moving configs for subtree sync

MikeWrock:


bgill92 (approved):
Checked to see if `picknik_ur_base_config` works with the reorg, and it still does, so LGTM

